PR #29, opened 2018-01-09
[Fix] Stops lockers from spawning more items each time the server enters a new round.

Kurgis:
Keeps the lockers, but stops spawning of items wherever the lockers are. Some lockers still spawn items:
*Russian Lockers
*Thunderdome Lockers
*HoP's Locker spawns a ID box and a 2 PDAs
*Doesn't touch wardrobes
*Anything with the New() tag instead of WillSpawn()


Kurgis:
Going to fix the spammed commits to keep the sanity of others.

PR #36, opened 2018-01-13
Removed code\modules\client\movement.dm

DukeDigby:
Removed the .dm containining the verbs for Spin CW/Spin CCW. It felt redundant unless you really do have a use for it.


DukeDigby:
Containing*. I should be more awake when i write.

XElectricX:
You do know people use those verbs, right?

DukeDigby:
First PR. I did mention in the PR Comment that it felt redundant, unless you had a use for it. Why people have a use for it is beyond me. Feel free to close.

PR #53, opened 2018-02-07
Circuits for the new buildable machines

XElectricX:
Added circuits to the imprinter, and a bed sheets pack from a while ago.


XElectricX:
Added many circuit boards to the imprinter, reworked the chameleon kit (more expensive, gun is separate), added in slime scanner to protolathe, and added gas and health analyzers to the autolathe.

PR #55, opened 2018-02-12
Buildable chem dispenser

XElectricX:
Allows the chem dispenser to be built.


XElectricX:
Crap, it was supposed to only contain the commit for the chem dispenser.

PR #97, opened 2018-04-18
Make closets and closet-type things anchorable

rhicora:
This includes closets, job closets, crates, etc. It does not include body bags or statues.


code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -15,6 +16,8 @@
 
 	var/welded = 0
 	var/large = 1
+	var/wrenchable = 1
+	var/unwrenched = 1 // Unwrenched closets can be anchored.
```
Zero-Bits: Why have another var when you could just check if the closet is anchored?

code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -363,6 +366,31 @@
 			playsound(src.loc, 'sound/weapons/blade1.ogg', 50, 1)
 			playsound(src.loc, "sparks", 50, 1)
 			open()
+	else if(isWrench(W))
+		if (src.wrenchable==0)
+			// Do not allow wrench interactions with things that aren't wrenchable.
+			return
+		if (src.unwrenched==0)
```
Zero-Bits: Same here. Why not if (src.anchored==1) ?

code/game/objects/structures/crates_lockers/closets.dm
Zero-Bits: Why have another var when you could just check if the closet is anchored?

code/game/objects/structures/crates_lockers/closets.dm
Zero-Bits: Same thing here. Wouldn't `if (src.anchored==1)` work?

code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -363,6 +366,31 @@
...
+			playsound(src.loc, 'sound/items/Ratchet.ogg', 50, 1)
+			to_chat(user, "<span class='notice'>You begin to unsecure \the [src] from the floor...</span>")
+			if (do_after(user, 40, src))
+				user.visible_message( \
+					"<span class='notice'>\The [user] unsecures \the [src].</span>", \
+					"<span class='notice'>You have unsecured \the [src]. Now it can be pulled somewhere else.</span>", \
+					"You hear ratchet.")
+				src.anchored = 0
+				src.unwrenched = 1
```
Zero-Bits: Another instance of src.unwrenched

code/game/objects/structures/crates_lockers/closets.dm
```diff
@@ -363,6 +366,31 @@
...
+		else /*if (src.unwrenched==1)*/
+			playsound(src.loc, 'sound/items/Ratchet.ogg', 50, 1)
+			to_chat(user, "<span class='notice'>You begin to secure \the [src] to the floor...</span>")
+			if (do_after(user, 20, src))
+				user.visible_message( \
+					"<span class='notice'>\The [user] secures \the [src].</span>", \
+					"<span class='notice'>You have secured \the [src].</span>", \
+					"You hear ratchet.")
+				src.anchored = 1
+				src.unwrenched = 0
```
Zero-Bits: And another.

PR #100, opened 2018-04-18
Fix the camera system to allow public and starter faction cams

rhicora:
What I'm trying to do:
- Add the 'nt_net' and 'freenet' security camera networks for each of the default factions.
- Require security program access (5) in the appropriate faction to access.
- Add a 'Public' network for public broadcasts.
- Add a 'Television' program which does nothing but consume Public cameras.

What I'm not trying to do:
- Add full camera support for custom factions.

### DIFF
--- a/code/modules/modular_computers/file_system/programs/generic/camera.dm
+++ b/code/modules/modular_computers/file_system/programs/generic/camera.dm
@@ -1,26 +1,20 @@
-// Returns which access is relevant to passed network. Used by the program.
-/proc/get_camera_access(var/network)
+// Returns whether a network is faction restricted
+/proc/is_faction_restricted(var/network)
 	if(!network)
 		return 0
-	. = GLOB.using_map.get_network_access(network)
+	. = GLOB.using_map.(network)
 	if(.)
 		return
 
 	switch(network)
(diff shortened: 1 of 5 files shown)


Zero-Bits:
Congrats on #100

code/modules/modular_computers/file_system/programs/generic/camera.dm
```diff
@@ -72,12 +76,17 @@
 /datum/nano_module/camera_monitor/proc/modify_networks_list(var/list/networks)
 	return networks
 
-/datum/nano_module/camera_monitor/proc/can_access_network(var/mob/user, var/network_access)
-	// No access passed, or 0 which is considered no access requirement. Allow it.
-	if(!network_access)
+/datum/nano_module/camera_monitor/proc/can_access_network(var/user, var/network, var/connected_faction, var/restricted)
```
Zero-Bits: `user` should always be a mob, and `var/mob/user` allows you to use some mob-specific procs and vars.

Zero-Bits: (If there's ever a case where `user` isn't a mob, something is going wonky-shaped)

PR #104, opened 2018-04-18
Major Slime Nerfs, Makes Slime Cores Orderable

Zero-Bits:
Metal, Adamantium, Black, Bluespace, and Dark Purple slimes can no longer be obtained via breeding.
Slimes are slightly less aggressive and take 6x longer to starve. They are also easier to wrestle.
Slimes have to eat 3x as much before evolving and splitting, and will only split into 2 instead of 4.
Slimes have had their mutation chance reduced to 5-15% instead of 25-35%
Slime health has been increased by 25%

Slimes should be able to go about 18-20 hours without eating now, but they will eat a lot more before growing. This means slimes will spread somewhat slowly and will be more of a passive activity.

Grey slime cores are orderable now, with a crate costing 50 and containing 2 cores. Inject a core with phoron to create a living grey slime. But be careful and probably wear a bio-suit.


Zero-Bits:
You'll want to make sure a slime's `nutrition` var is saved when the server saves and loads by the way.

PR #106, opened 2018-04-18
Virology machines, library machines, orderable slime cores, mining machines

XElectricX:
The commits already explain what is what.


code/game/objects/structures/mop_bucket.dm
```diff
@@ -25,3 +25,8 @@
 			reagents.trans_to_obj(I, 5)
 			to_chat(user, "<span class='notice'>You wet \the [I] in \the [src].</span>")
 			playsound(loc, 'sound/effects/slosh.ogg', 25, 1)
+	if(isWrench(I))
+		playsound(src.loc, 'sound/items/Ratchet.ogg', 75, 1)
+		to_chat(user, "<span class='notice'>You deconstruct the [src]/span>")
```
Zero-Bits: `"<span class='notice'>You deconstruct the [src]/span>"` is missing an `<` on that closing span tag. That's what's Travis is kicking you about.

code/modules/library/lib_machines.dm
```diff
@@ -9,17 +9,31 @@
  * Library Scanner
  */
 /obj/machinery/libraryscanner
-	name = "scanner"
+	name = "Book Scanner"
```
Zero-Bits: should be "\improper Book Scanner" or "book scanner" otherwise you'll get some wonky stuff on examine.

Zero-Bits: Otherwise, everything looks good.

code/game/objects/structures/mop_bucket.dm
Zero-Bits: Should also be `\the [src]` rather than `the [src]`

PR #107, opened 2018-04-20
Supply Shuttle Fix

Zero-Bits:
Quick patch to supply shuttles. They'd toss a runtime if the path for their thing didn't exist. This would cause Travis to fail on unrelated PRs like #106


Zero-Bits:
This is the last PR I'll have time to make before I leave for Texas, but I'll likely still be able to diagnose PRs that fail Travis and provide simple reviews.

PR #111, opened 2018-04-22
Added eject into hand to many devices

robbym:
* Changed many devices that eject something to eject into the user's hand if it is empty.


code/modules/reagents/Chemistry-Machinery.dm
```diff
@@ -19,7 +19,7 @@
 	idle_power_usage = 20
 	clicksound = "button"
 	clickvol = 20
-	var/beaker = null
+	var/obj/item/weapon/reagent_containers/beaker  = null
```
Zero-Bits: There's a double space between `beaker` and `=`

PR #112, opened 2018-04-22
Ports Hydrogen additions from Bay and adds Stasis clamp to the autolathe

NataKilar:
-Hydrogen filtering adapted/ported from Bay for use in power generation and atmospherics. Unique sprite added to hydrogen cans
-Stasis clamps added to the autolathe


Zero-Bits:
Hi. I did all this hydrogen stuff for bay in the first place. One important thing to consider is that you cannot detect hydrogen right now.

Zero-Bits:
You'll need to port another bay PR for that.

Zero-Bits:
You should probably port this as well: https://github.com/Baystation12/Baystation12/pull/20631

PR #115, opened 2018-04-22
Add colored light tubes and colored light bulbs to supply

rhicora:
They work the same as normal lights, but with a different color. Ideal for your rave lair!


rhicora:
For some reason this got the code from https://github.com/Persistent-SS13/Persistent-Bay/pull/114 so merge that first.

PR #118, opened 2018-04-23
Berets, Earrings, Journalism, Law

rhicora:
- Adds the Beret Crate, for ordering lots of luxury berets. So soft and French!
- Adds the expensive, possibly jewelry-filled Luxury Crate
- Adds the Journalist Crate, for budding correspondents
- Removes access checks for holobadges


Zero-Bits:
Code still looks good. Still probably passes Travis.

Zero-Bits:
Yep. Code still checks out, and I've no doubts that it will pass Travis.

PR #121, opened 2018-04-24
Patch for Televisions / Cameras

rhicora:
This patches some issues with Television and Camera Monitor.

- Public network is now viewable on Camera Monitor in addition to Television
- Press camera drones now broadcast to public channel
- Software list can be refreshed by turning NTNet downloads off and on again

### DIFF
--- a/code/game/objects/items/devices/tvcamera.dm
+++ b/code/game/objects/items/devices/tvcamera.dm
@@ -6,7 +6,7 @@
 	w_class = ITEM_SIZE_LARGE
 	slot_flags = SLOT_BELT
 	var/channel = "General News Feed"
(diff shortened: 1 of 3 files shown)


code/game/objects/items/devices/tvcamera.dm
```diff
@@ -6,7 +6,7 @@
 	w_class = ITEM_SIZE_LARGE
 	slot_flags = SLOT_BELT
 	var/channel = "General News Feed"
-	var/obj/machinery/camera/network/thunder/camera
+	var/obj/machinery/camera/network/public/camera
```
Zero-Bits: Some day I'd like to be able to pick the network my drone broadcasts on. Some day.

PR #122, opened 2018-04-24
Pay Refugees

rhicora:
This code change causes refugees and other factions to be paid. They will get back pay for the time worked since that will have been logged as "worked" in their crew record.

982a513 caused payday timer to advance by one hour immediately upon issuing pay. Since NT was the first faction in the list, this was absolutely no problem for them, but by the time Refugee was evaluated, it had already become "not payday yet."


FaurnsGoldberg:
ROFLing at the problem, nice work!!

PR #126, opened 2018-04-28
Mineral Sublimation

NataKilar:
Adds a portable machine (like a pump or canister) that can convert mineral sheets into gas. Numbers are tentative, and generation for ores is still needed.


code/modules/materials/materials.dm
```diff
@@ -628,6 +632,14 @@ var/list/name_to_material
 	stack_origin_tech = list(TECH_MATERIAL = 6, TECH_POWER = 6, TECH_MAGNET = 5)
 	is_fusion_fuel = 1
 
+/material/ice
+	name = "ice"
+	stack_type = /obj/item/stack/material/ice
+	icon_colour = "#c5e5e2"
+	gaseous_products = list(
+				"hydrogen" = 22.2, //(600g / 18 grams per mole)*2/3
+				"oxygen" = 11.1
```
Zero-Bits: This will mix and instantly form water vapor.

Zero-Bits: Not requesting changes, just making a comment. Maybe someday we'll have a water separator.

Zero-Bits:
Looks good so far. I can confirm that the Travis error is an issue with shuttles, not with this pr in specific.

code/modules/mining/drilling/drill.dm
```diff
@@ -29,7 +29,9 @@
 		"hydrogen" = /obj/item/weapon/ore/hydrogen,
 		"silicates" = /obj/item/weapon/ore/glass,
 		"carbonaceous rock" = /obj/item/weapon/ore/coal,
-		"bluespace crystal" = /obj/item/bluespace_crystal
+		"bluespace crystal" = /obj/item/bluespace_crystal,
+		"ice" = /obj/item/weapon/ore/ice/,
+		"dryice" = /obj/item/weapon/ore/dryice/
```
Zero-Bits: Not really your issue, but I'd appreciate it if you could add `"rock salt" = /obj/weapon/ore/salt/`

code/modules/random_map/noise/ore.dm
```diff
@@ -83,6 +85,8 @@
 				T.resources["diamond"] =  0
 				T.resources["iron"] =     0
 				T.resources["salt"] =     0
+				T.resources["ice"] = 	  0
+				T.resources["dryice"] =   0
```
Zero-Bits: You'll also need to add ice and dry ice to `code/modules/mining/mine_turfs.dm` or it won't generate.

PR #130, opened 2018-04-29
More autolathe recipes

rhicora:
Adds autolathe recipes for the following: pitcher, carafe, coffee cup (all types), spray bottle, dropper, ecigarette, beer keg, water cooler. Fixes inconsistent capitalization in autolathe recipes.

### DIFF
--- a/code/game/machinery/autolathe_datums.dm
+++ b/code/game/machinery/autolathe_datums.dm
@@ -550,6 +550,16 @@ var/const/EXTRA_COST_FACTOR = 1.25
 	path = /obj/machinery/clamp
 	category = "Engineering"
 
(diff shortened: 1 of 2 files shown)


code/game/machinery/autolathe_datums.dm
```diff
@@ -111,7 +111,7 @@ var/const/EXTRA_COST_FACTOR = 1.25
 	category = "Tools"
 
 /datum/autolathe/recipe/t_scanner
-	name = "T-ray scanner"
+	name = "t-ray scanner"
```
Zero-Bits: It would actually be T-ray, since the T stands for something.

PR #131, opened 2018-04-29
Add bulk clothes crate and cigar crate

rhicora:
Adds a bulk clothes crate (x10 random clothing or jacket) and a worryingly expensive imported cigar crate. Also lets security specifically order orange jumpsuits.


robbym:
Nope. https://github.com/Persistent-SS13/Persistent-Bay/issues/132

PR #134, opened 2018-04-29
Made meat spikes (de)constructable

robbym:
Made meat spikes (de)constructable.


robbym:
Needs https://github.com/Persistent-SS13/Persistent-Bay/pull/133 to compile.

PR #138, opened 2018-04-30
Bugfixes for Clothes Crates and Autolathe Recipes

rhicora:
Fixes some recently introduced bugs.


code/game/machinery/autolathe_datums.dm
```diff
@@ -674,10 +674,6 @@ var/const/EXTRA_COST_FACTOR = 1.25
 /datum/autolathe/recipe/coffeecup/rainbow
 	path = /obj/item/weapon/reagent_containers/food/drinks/coffeecup/rainbow
 
-/datum/autolathe/recipe/coffeecup/tall
-	path = /obj/item/weapon/reagent_containers/food/drinks/coffeecup/tall
-	hidden = 1 // Yuki can't be trusted with this.
```
Zero-Bits: I am hurt by your choice to remove my greatest work.

code/modules/clothing/under/accessories/clothing.dm
```diff
@@ -188,6 +188,15 @@
 	var/tucked = 0
 	var/buttoned = 0
 
+/obj/item/clothing/accessory/toggleable/flannel/red
+	name = "red flannel shirt"
+	desc = "A comfy, red plaid flannel shirt."
+	icon_state = "flannel"
+	var/rolled = 0
+	var/tucked = 0
+	var/buttoned = 0
```
Zero-Bits: Remove `var/` from these three vars, since they were already defined in the parent object.

code/game/machinery/autolathe_datums.dm
rhicora: Are you the infamous creator of Tall Coffee Cup?

Zero-Bits: Yes. I wasn't aware that I was infamous though.

PR #142, opened 2018-05-02
Label Removal

NataKilar:
Allows the removal of labels by clicking again with an active hand labeler


code/modules/paperwork/handlabeler.dm
```diff
@@ -49,6 +52,13 @@
 		spawn(1)
 			tray.update_icon()
 
+	if(findtext(A.name, "(")) //Check if the item is already labeled
+		A.name = copytext(A.name, 1, findtext(A.name, "(")-1) //Remove any labels
```
rhicora: Will this also remove the space prior to the open paren?

NataKilar: Yep, that’s the -1 after the findtext function

rhicora:
Looks great.

PR #144, opened 2018-05-02
Buildable secure crates, fixes to air tank dispenser, buildable ore box, jetpacks, printing paper, etc.

XElectricX:
Tons of small changes, which is why they are being lumped into 1 PR. The committs show each change, but the main ones are jetpacks can be made at a protolathe, miners get the right scanner (they used to only have the xenoarcheology scanner), the ore scanner has a new sprite, paper can be made at the biogenerator, and air tanks made at the autolathe do not spawn with air.


code/game/objects/structures/crates_lockers/crates.dm
```diff
@@ -342,6 +342,16 @@ obj/structure/closet/crate
 	icon_opened = "largermetalopen"
 	icon_closed = "largermetal"
 
+/obj/structure/closet/crate/secure/large/reinforced/proc/slice_into_parts(obj/item/weapon/weldingtool/WT, mob/user)
```
Zero-Bits: Should just be `slice_into_parts| rather than `proc/slice_into_parts`

The proc is already initially defined in a parent of the object, so it only needs to be referenced.

XElectricX: Ooooh, alright.

Zero-Bits:
```
[07:00:00] Runtime in autolathe_datums.dm,72: Cannot modify null.starting_pressure.
```

XElectricX:
Damn indentations!

FaurnsGoldberg:
you cant just change things like spider balance with zero discussion, and unfortunately thats lumped with a bunch of changes i didnt even read. please ammend.

PR #147, opened 2018-05-03
Adds options for turrets and buffs them slightly. Fixes sublimator typo

NataKilar:
Adds range option for turret, with maximum range remaining the same as current
Buffs turret health
Turrets now require plasteel to construct.


Zero-Bits:
(referring to turrets already in the game)

PR #169, opened 2018-05-08
Added peeking (alt + dir), with a rebound faceperm (alt + shift + dir)

PrPleGoo:
Alt+WASD in hotkey mode allows you to "throw" your eye (arrows/numpad also work).
Allows 5 extra turfs of sight in the chosen direction.
Works for diagonals too.
Should obey rules of visibility as that's done from the PoV from your mob.
Rebound faceperm to alt+shift+arrows/numpad.

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -1006,7 +1006,7 @@ mob/proc/yank_out_object()
 	. = stat != new_stat
 	stat = new_stat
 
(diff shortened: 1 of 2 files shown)


FaurnsGoldberg:
faceperm should remain the same and its the peeking that should use alt+shift+direction

PrPleGoo:
@FaurnsGoldberg you are wrong

PR #170, opened 2018-05-08
Added peeking (shift + dir / alt + WASD)

PrPleGoo:
Shift+WASD in hotkey mode allows you to "throw" your eye (arrows/numpad also work).
Allows 5 extra turfs of sight in the chosen direction.
Works for diagonals too.
Should obey rules of visibility as that's done from the PoV from your mob.


FaurnsGoldberg:
Not happening until faceperm is unchanged

FaurnsGoldberg:
well alright then

PR #174, opened 2018-05-08
Removes Torch References, Removes Mapmerge, and adds Nullspace Test

Zero-Bits:
The nullspace test essentially makes sure mobs don't throw runtimes when put into nullspace.
We don't use mapmerge because we don't edit the map more than once in a blue moon.
We also don't use the Torch.


Zero-Bits:
Right. It times out because our map is too big. RIP.

PR #176, opened 2018-05-08
Revert girderlathe + wall changes

rhicora:
After talking in #code-suggestions, we want to propose reverting girderlathe changes for now. Some of the potential flaws that were brought up in that channel:

- It makes more sense that the girders of a space station would be assembled on-site.
- Normal station construction should not be dependent on yet another science board.
- We already have a way to slow down girder construction; increase its timer.
- There is a strong non-tide player demand for false walls.
- The girderlathe is flawed in that it cannot be moved once built.


Zero-Bits:
Don't forget to update globals and Travis.

Zero-Bits:
I'd recommend holding on a merge until discussion has actually led to a conclusion.

rhicora:
Talked with John Anderson and this is a Brawler decision. I'm still concerned Girderlathe doesn't have any quality of life benefits to outweigh its costs.

rhicora:
Closing this for now, reopen if needed.

PR #180, opened 2018-05-09
Reworks Material Grinding

Zero-Bits:
Materials can now be ground into multiple reagents and can have different amounts of each reagent per sheet.

Kept the old grinding recipes, and also added a few more:
Ice to water and ice
Dry ice to carbon
Iron to iron
Steel to iron and carbon
Plastic to plasticide
Wood to a little carbon
Phoron-glass to a little phoron


Zero-Bits:
Gave Travis a kick in the boot.

FaurnsGoldberg:
Looks good, travis passed

PR #181, opened 2018-05-09
Makes Time Configurable, Links IC time to realtime

Zero-Bits:
Year skip and Timezone are config options. Roundstart time is now equal to the current time GMT +- timezone hours. 

Year skip default has been set to 420 years, down from baycode's 544, and up from our lore's 400. It's 420 because that syncs up perfectly with the Gregorian calendar's cycle, so the calendar will be exactly the same as the real current year's.

Timezone default has been set to EST (GMT-5) because /shrug

### DIFF
--- a/code/_helpers/global_access.dm
+++ b/code/_helpers/global_access.dm
@@ -823,8 +823,8 @@
 			return global.round_progressing;
 		if("round_start_time")
 			return global.round_start_time;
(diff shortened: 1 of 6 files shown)


FaurnsGoldberg:
wow this is great, reviewed and it looks fine. hopefully we can implement more timestamps in game, as this really works with persistence

PR #182, opened 2018-05-09
Makes Us 512 Compatible

Zero-Bits:
Replaces some (reserved by 512) keywords to make it so we don't immediately fail to compile for 512.


FaurnsGoldberg:
Looks good. I was worried about 512 compatibility and i don know the first thing about it.

PR #187, opened 2018-05-09
Expands Mining and Materials

Zero-Bits:
Adds some new mine-able minerals and corresponding materials:
Copper
Aluminum
Sulfur
Lead
Tungsten
Tin
Zinc

Copper, Aluminum, Sulfur, and Tungsten can be ground into their corresponding chemistry reagents.

Tin and Zinc can be alloyed with Copper for Bronze and Brass respectively.

Also adds all new materials (including ones from previous PRs) to the item_worth sheet.

This should provide a more renewable way to get a few more base chemistry reagents.

Coal is also compressible now, and coal bricks can be put in a grinder for a rich source of carbon.


Andrew-Fall:
I personally would like to see us avoid using parents for materials, unless it is a very generic thing, such as /material/metal
So instead of /material/copper/brass, I think they should be separate, /material/copper, /material/brass

Zero-Bits:
@Andrew-Fall no more parenting. The Travis error is due to your wall stuff being reverted, but not your travis stuff.

Andrew-Fall:
Looks good

PR #188, opened 2018-05-10
Cold Medicine Lasts Slightly Longer. Also, Stasis Has Been Completely Overhauled.

Zero-Bits:
Cold medicine is no longer such a hassle. It lasts a decent while now.

Less importantly, stasis has been entirely overhauled.

Stasis is now a factor for life ticks instead of binary on/off thing. E.g with stasis 2, half of ticks will be skipped without processing and so on.
Stasis can now have different sources, similar to chem effects.
Being very cold applies stasis, depending on species temperature thresholds.
Being in stasis applies drowsiness, high drowsiness makes you fall asleep.

Cryo no longer multiplies all inejcted chems by 10. Cryochems metabolism adjusted to compensate.

Cryochems once again repair organs, but they're capped in the same way as peri. (i.e. can only heal minor organ wounds)

Cryopod's protection against extreme cold is now a chemical effect of cryochems.

Sleepers have stasis, none/2x/5x settings.

Cryobags start at 20x stasis, and deteriorate every 5 minutes by 25%, turning useless after 40 minutes. Examine messages and visual indication added for their power.

Stolen from the following Bay PRs:
https://github.com/Baystation12/Baystation12/pull/20114 
https://github.com/Baystation12/Baystation12/pull/20239 
https://github.com/Baystation12/Baystation12/pull/20243 
https://github.com/Baystation12/Baystation12/pull/20253 
https://github.com/Baystation12/Baystation12/pull/20268 
https://github.com/Baystation12/Baystation12/pull/20802

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
@@ -157,10 +157,17 @@
 	if(M.bodytemperature < 170)
 		M.adjustCloneLoss(-100 * removed)
 		M.add_chemical_effect(CE_OXYGENATED, 1)
-		// It metabolizes 10x slower, but only heals 2x as much.
-		// This means that it heals at 1/5 the speed it used to.
-		// Stasis further slows it.
-		M.heal_organ_damage(20 * removed, 20 * removed)
+		if(ishuman(M))
+		var/mob/living/carbon/human/H = M
+		for(var/obj/item/organ/internal/I in H.internal_organs)
+			if(I.robotic >= ORGAN_ROBOT)
+				continue
+			if(I.organ_tag == BP_BRAIN)
+				H.confused++
+				H.drowsyness++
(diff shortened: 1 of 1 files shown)


Zero-Bits:
Travis is still only failing due to the wall stuff. This PR doesn't trigger anything.

Andrew-Fall:
It looks like stasis_settings is more or less useless. I don't believe you need to do that check, or create a variable for it

Zero-Bits:
@Andrew-Fall stasis_settings is the list of "settings" on the sleeper. (i.e. you can pick no stasis, 2x stasis, or 5x stasis)

Andrew-Fall:
Yes, but seeing as it is handled by the template, and only used to check input received with the template, which is hard-coded, it appears redundant.

Zero-Bits:
Actually, it's necessary to prevent exploits. It's not that hard to change the client-contents of an HTML window. `if(stasis != nstasis && nstasis in stasis_settings)` ensures that the chosen setting actually matches the available settings.

PR #194, opened 2018-05-10
Peek remap, adds control+wasd

matt81093:
So this changes peeking to shift+alt+dir and adds control+wasd for pointing your char in a direction. I grouped the keys when I was going through them to find stuff easier although the diff looks terrible.

I would use a :cl: but I'm a bit tired to put something there. I need to test this to make sure I didn't accidentally delete something.

### DIFF
--- a/interface/skin.dmf
+++ b/interface/skin.dmf
@@ -1,224 +1,3 @@
-macro "borghotkeymode"
-	elem 
-		name = "Tab"
-		command = ".winset \"mainwindow.macro=borgmacro hotkey_toggle.is-checked=false input.focus=true input.background-color=#D3B5B5\""
-	elem 
-		name = "Center+REP"
-		command = ".center"
-	elem 
-		name = "Northeast"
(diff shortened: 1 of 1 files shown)


Zero-Bits:
Shift+Dir for peeking would be more consistent with what brawler decided on. Is there a reason you're changing that?

matt81093:
Shift+dir is used by some for other personal macros. Curious about Shift+Ctrl though. Could also remove it entirely, but the issue then becomes people not knowing about it, or those who don't use personal macros. 
I mainly wanted this because modifier keys weren't being used consistently for the peek function.

matt81093:
Erm. Until peeking is fixed, I might remove the macro @Zero-Bits It's apparently causing issues

Zero-Bits:
@matt81093 Righto. I'll wait to hear back from you on that.

matt81093:
Closing this since peeking was removed

PR #201, opened 2018-05-11
Remove unnecessary undefines

XElectricX:


Zero-Bits:
This is waiting on the new pipe dispenser recipes.

PrPleGoo:
check the object tree for all pipes

XElectricX:
Trying to add a simple gas injector has been taking up too much of my time so I'll just leave this as is. There are more important things that need to be fixed.

PrPleGoo:
Undeffing these is to prevent the constants from being used anywhere else for any other purpose than pipe construction.

FaurnsGoldberg:
If the defines are only used within the file, undefines seem appropriate to me.

PR #206, opened 2018-05-12
Welding gas mask, bugfix to quick-equip, readdition of peek (unbound), clicking on inventory slots

PrPleGoo:
Added welding gas mask.
It's a gas mask that can also protect you from damaging your eyes.
It has a toggle like the welding helmet.
Doesn't protect you from eye-stabbing in the upright position.
Fixed an inventory proc that used some retarded proc instead of the intended bag proc (quick-equip called this resulting in retarded pointer behaviour).
Added peek back in but didn't bind it to any hotkeys, leaving it up to players to use it.
Clicking on inventory slot boxes inside bags will instead interact with the item.


FaurnsGoldberg:
the title for your PR should be better

code/modules/mob/mob.dm
```diff
@@ -1020,6 +1020,38 @@ mob/proc/yank_out_object()
 	set hidden = 1
 	set_face_dir(client.client_dir(WEST))
 
+/mob/verb/eastfacepeek()
```
Zero-Bits: Undocumented re-addition of peek.

code/modules/mob/inventory.dm
```diff
@@ -87,20 +87,26 @@ var/list/slot_equipment_priority = list( \
 
 	return 0
 
-/mob/proc/equip_to_storage(obj/item/newitem)
+/mob/proc/equip_to_storage(obj/item/newitem, var/force = 1)
```
HeyBanditoz: You should document the reasoning for `force` with a comment. You will also have to update `equip_to_storage()` in [these places.](https://github.com/Persistent-SS13/Persistent-Bay/search?utf8=%E2%9C%93&q=equip_to_storage&type=)

PrPleGoo: No I don't since the default is set to 1 it can be called without a 2nd argument.

code/modules/mob/mob.dm
PrPleGoo: Added to desc.

PR #207, opened 2018-05-12
Fixes Autolathed Tanks

Zero-Bits:
They now actually spawn empty. Cargo-ordered tanks should spawn full. If they don't, then that's another issue altogether.


Zero-Bits:
Also went ahead and added an /empty type to all tanks, just in case an admin needs to spawn one in empty.

Zero-Bits:
Don't merge this until #208 goes through. This will have less conflicts to deal with.

Andrew-Fall:
The issue was that fabricate was never called, autolathes should be instead changed to act like protolathes

Zero-Bits:
This is being closed to prevent conflicts with #211

PR #208, opened 2018-05-12
Reworks Autolathe, Adds Autolathe Recipe Tests to Travis

Zero-Bits:
The autolathe can now take any stacked material and print with any material. There is also a couple new Travis tests that ensures all autolathe recipes are valid.

To showcase this, I've added some circuits that take various precious metals (copper, silver, gold, platinum); the kind of stuff you'd need to start up a station before you can get an R&D department running.

I eventually plan to port over most handcrafted recipes to the autolathe, and give it a category system that isn't garbage.

### DIFF
--- a/code/_helpers/global_access.dm
+++ b/code/_helpers/global_access.dm
@@ -773,6 +773,10 @@
 			return global.prometheans;
 		if("protected_objects")
 			return global.protected_objects;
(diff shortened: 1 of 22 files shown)


Andrew-Fall:
Looks good, however, see #207

PR #209, opened 2018-05-12
Racks shouldn't yield free plastic

rhicora:
Make the default table material steel, and yield that when disassembling racks.


Andrew-Fall:
Why change default table material when you can change the rack to just be steel?

PR #216, opened 2018-05-15
Comments out game-breaking boards

UndeadNappist:
Both holocasters and exosuits break the game, and telescience is defunct.
The exosuit fabricator was left in because it has valuable research items on it.


FaurnsGoldberg:
Mechas do cause problems, but I dont think commenting out their construction is what we need to do now.  I believe its only loaded mechas that cause it, and I believe that we will eventually (hopefully soon) have that issue worked out so that mechas can be a big part of the game again. I also dont know how holocasters break the game, I know they havent been working right all the time, but I've seen them work and they are very cool.

UndeadNappist:
I agree with you, Faurns. That is why I commented them out instead of deleting them, and why cyborgs were left untouched despite also being defunct - nobody builds them and it doesn't crash the active game and force rollbacks, deleting people's hard in-game work.
Whenever these things are fixed either comment them back in or revert this PR.

PR #217, opened 2018-05-15
Save improvement Part 1/3, Saving Records as its own file

FaurnsGoldberg:
In an effort to reduce RAM use, I'm going to be trying a number of things to split the files. This makes records and factions stored in a records.sav file, I've already tested the full system, but I need to merge it in three parts because I need to first

Have the server load like it used to, but save with the new files generated, then

Merge the other half and restart, so that the load now uses the new files and not the old save, and finally

Make the final change which takes the old unnecessary save protocols out, this is saved for last so that any files created prior to this are valid backups.


FaurnsGoldberg:
it fails travis because i do a <<!! this is no good, travis shouldnt even check for that (imo)

FaurnsGoldberg:
well i fixed it, by leaving a part commented out which should have been commented out anyways, for this part (although it wouldnt have hurt anything anyways)

PR #227, opened 2018-05-22
Nice autolathe + sane recycler

PrPleGoo:
Added multiplier to machinery. 
Implemented multiplier for autolathe. 
Fixed wire coil recipe.
Fixed poor math on autolathe recipe. 
Fixed autolathe using an ancient window method. 
Allowed autolathe to take any material. 
Buffed autolathe storage and crafting speed.

Recycler now eats materials of items and spits out sheets of the correct stuff. 
Recycler doesn't recycle any type of stack.

Increased default item spread.

### DIFF
--- a/code/game/machinery/autolathe.dm
+++ b/code/game/machinery/autolathe.dm
@@ -184,10 +184,8 @@
 			var/obj/item/stack/stack = eating
 			total_material *= stack.get_amount()
 		else
-			world << "A total_material: [total_material]"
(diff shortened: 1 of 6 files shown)


PrPleGoo:
839 << uses (expecting exactly 840)
Travis doesn't like improvements :(

Zero-Bits:
In `test\check-paths.sh` change 840 to 839

Zero-Bits:
Also, since the recycler doesn't recycle stacks, that means it can't recycle cable coils.

PR #228, opened 2018-05-24
Vintage .45 fits in security belt, biogenerator prints paper sheets

XElectricX:
Vintage .45 pistol, like the regular one, now fits in the security belt. Changed the biogenerator to print paper sheets instead of empty bundles.


Kurgis:
Why not have the Biogenerator produce cardboard instead, and then craft paper from the cardboard? This would mean you wouldn't have 5 unstacked items being made on top of the biogenerator at once.

PR #238, opened 2018-05-26
Faction creation fix

### LINKED ISSUE #237: [BUG] Bluespace Satellites require no components
#### Description of issue
You can create a Bluespace Satellite without needing to use any Telecommunications parts, or a bluespace crystal as it says in the code. This is because the circuit board does console frame construction instead of machine frame construction, I'm unable to tell why this is occurring. It's a problem since factions can be created with an incredibly low material cost with only the board required.

This went unreported for awhile since players assumed it was intended behavior.
I expected to need to place the Bluespace Satellite board into a machine frame, and then place the required components to create a bluepsace satellite.

Instead I place the circuit board into a computer frame, use some wire and glass, and I get a bluespace satellite.
#### Steps to reproduce 

1. Print Bluespace Satellite board from Circuit Imprinter.
2. Attempt to create faction by placing into machine frame.
3. Machine frame denies board.
4. Place into computer frame.
5. Get Bluespace-free Bluespace Satellite.

https://github.com/Persistent-SS13/Persistent-Bay/blob/43db64731fc04c5384b33ab924902885ce6a6a21/code/modules/factions/bluespace_satellite.dm#L195
https://github.com/Persistent-SS13/Persistent-Bay/blob/32d5aebd567f2829e31e51422dec6ac3e7b91709/code/game/objects/items/weapons/circuitboards/machinery/telecomms.dm#L18

It seems that this has always been happening since they were made to be craftable through R&D, no one has even seen a Bluespace Crystal in-game yet.

Server Revision: 5f88f7ba7839383c3cd849182fece2683ae5ff29 - dev

#### Issue bingo

- [ ] Issue could be reproduced at least once
- [ ] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/Persistent-SS13/Persistent-Bay/issues/)

Kurgis:
Fixes #237
Was missing a small /telecomms in the item's path, causing the circuit board to default to console construction.


Zero-Bits:
I've restarted the Travis build for you.

PR #239, opened 2018-05-26
Fixes Travis Error

Zero-Bits:
#232 caused an error with Travis. This fixes that.


Zero-Bits:
Now this also adds a little bit of code that will _hopefully_ prevent future incidents of this nature.

PR #243, opened 2018-05-28
Reworked Advanced Energy Gun

XElectricX:
Introducing the HC30 Laser Carbine! I modified the AEG from a self-charging energy gun into a variant of the standard laser rifle, but with more ammo. Would have been easier to make the gun a child of the laser gun, but the AEG itself had to be modified for changes to take effect. Will make it its own individual item during the next server wipe and dump the existing AEG. The gun has also been made more expensive.


Zero-Bits:
This PR must be changed so that it doesn't conflict.

FaurnsGoldberg:
The AEG has been changed in #248

Kurgis:
I do not think it's an appropriate fix, what XElectricX did was remove the nuclear gun's code, and effectively replace it with a laser gun clone. The gun is three laser gun charges with no downsides, yes, he did make an updated sprite (without in-hands), but it doesn't make sense to make the gun share the same functionality of another. The purpose of the NucGun was that it'd recharge, the purpose of the laser gun is that'd it be killy while having a good ammo count, the laser accelerator is extra killy with low ammo count, etc. This PR completely replaces the laser gun in all purpose, while having absolutely zero downsides unlike other higher-tier weapons like the laser accelerator. If that is the only proposed solution to increase max-charge, I'd suggest completely removing the NucGun since the laser gun fills the same exact purpose.

PR #250, opened 2018-05-29
Fixes Multi-Organ Bug

Zero-Bits:
You will no longer be able to attach two hearts, eight kidneys, forty lungs, etc.


FaurnsGoldberg:
this is good, but i think organ redundancies should be looked at as a feature

PR #253, opened 2018-05-30
Update lungs.dm

UndeadNappist:
See d672921

### DIFF
--- a/code/modules/organs/internal/lungs.dm
+++ b/code/modules/organs/internal/lungs.dm
@@ -20,7 +20,7 @@
 	var/min_breath_pressure
 	var/last_int_pressure
 	var/last_ext_pressure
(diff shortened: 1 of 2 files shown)


code/modules/organs/internal/lungs.dm
```diff
@@ -20,7 +20,7 @@
 	var/min_breath_pressure
 	var/last_int_pressure
 	var/last_ext_pressure
-	var/max_pressure_diff = 60 //Default
+	var/max_pressure_diff = 202.65 //Default
```
PrPleGoo: can you define this as two atms?

UndeadNappist:
Explicitly defined, as twice requested.

Zero-Bits:
Continues to look good and pass Travis :+1:

PR #255, opened 2018-06-03
90% of the shuttle update

FaurnsGoldberg:
Everything except them moving, I'll do that soon

### DIFF
--- a/code/game/shuttle_engines.dm
+++ b/code/game/shuttle_engines.dm
@@ -36,7 +36,7 @@
 /obj/structure/shuttle/engine
 	name = "engine"
 	density = 1
-	anchored = 1.0
+	anchored = 0.0
 	var/permaanchor = 0
 /obj/structure/shuttle/engine/verb/rotate()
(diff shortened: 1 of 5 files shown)


code/game/shuttle_engines.dm
```diff
@@ -36,7 +36,7 @@
 /obj/structure/shuttle/engine
 	name = "engine"
 	density = 1
-	anchored = 1.0
+	anchored = 0.0
```
Zero-Bits: Why is this a decimal float? IIRC anchored is always either 0 or 1.

code/modules/factions/bluespace_satellite.dm
```diff
@@ -107,6 +107,7 @@ GLOBAL_LIST_EMPTY(all_docking_beacons)
 	if(href_list["status_close"])
 		if(check_obstructed())
 			status = 5
+			status = 5
```
Zero-Bits: The line right above this one already sets status to 5.

PR #260, opened 2018-06-05
You Can Now Only Choose Factions With Active Beacons in Character Creation

Zero-Bits:
Also fixes that annoying as hell "Runtime in storage.dm,345: Cannot create objects of type /datum/storage_ui/default" spam.


code/modules/client/preference_setup/general/05_background.dm
```diff
@@ -88,7 +83,15 @@
...
 	else if(href_list["faction"])
-		var/list/joinable = list("Nanotrasen", "Refugees", "Entrepreneur")
+		var/list/joinable = list("nanotrasen")
+
+		if(!GLOB.frontierbeacons.len)
+			message_admins("WARNING! No beacons avalible for faction selection! spawn one and set the req_access_faction!")
+		for(var/obj/structure/frontier_beacon/beacon in GLOB.frontierbeacons)
+			if(!beacon.loc || !beacon.activated) continue
+			if(beacon.req_access_faction && !(beacon.req_access_faction in joinable))
+				joinable += beacon.req_access_faction
```
FaurnsGoldberg: This should be joinable |= beacon.req_access_faction to prevent NT from getting a duplicate

Zero-Bits: Currently, that's solved by `!(beacon.req_access_faction in joinable)` but your solution seems better.

Zero-Bits:
Don't merj this. It's not done.

PrPleGoo:
>hell
Please clean up your language in the PR.

Zero-Bits:
Heck you! I do whatever the flip I want! This ain't no danged Christian server!

Zero-Bits:
Right. Should be ready to roll. Just make sure to set "activated" to save on the beacons.

Zero-Bits:
@PrPleGoo It still gives a runtime on picking a faction. Something about a bad client.

Zero-Bits:
Don't merge this. The faction picker doesn't actually let you pick a faction and gives off a runtime when you try.

PR #263, opened 2018-06-10
Lots of shuttle things to test..

FaurnsGoldberg:
should be everything required to build a shuttle and even launch it, but i want to test it on the main serv


Zero-Bits:
Was pretty busy yesterday. I'll try to get reviews squared away today after work.

PR #269, opened 2018-06-15
shuttle polish part 1

FaurnsGoldberg:
shuttles should load properly now, cleans up the menu slightly, adds a bit of warmup time that will hopefully let the sound play

### DIFF
--- a/code/modules/factions/bluespace_satellite.dm
+++ b/code/modules/factions/bluespace_satellite.dm
@@ -9,9 +9,13 @@ GLOBAL_LIST_EMPTY(all_docking_beacons)
 	plane = LIGHTING_PLANE
 	layer = LIGHTING_LAYER
 	should_save = 0
+	
+	
 /obj/structure/hologram/dockzone
 	icon = 'icons/obj/machines/dock_beacon.dmi'
 	icon_state = "dockzone"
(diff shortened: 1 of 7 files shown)


code/modules/factions/bluespace_satellite.dm
```diff
@@ -90,7 +102,7 @@ GLOBAL_LIST_EMPTY(all_docking_beacons)
 			if(5)
 				data["status"] = "Obstructed"
 		data["dimenson"] = dimensions
-
+	message_admins("debug 1")
```
Zero-Bits: This seems incredibly vague. It should probably be made more specific for future debugging.

code/modules/factions/bluespace_satellite.dm
```diff
@@ -101,6 +113,7 @@ GLOBAL_LIST_EMPTY(all_docking_beacons)
 		ui.set_initial_data(data)
 		// open the new ui window
 		ui.open()
+		message_admins("debug 2")
```
Zero-Bits: This also seems really vague, and should probably show a message mentioning what's actually supposed to be going on.

Zero-Bits:
Still looks good :+1:

PR #276, opened 2018-06-19
[DNM] Contour Colony

Kurgis:
There may be loose tags in some items, hopefully Travis is set-up to find them. No, the Colony shown likely actually won't be the starter station after a conversation I had with Zero, and it's possible that the starter stations will be made more static.


Zero-Bits:
The Travis error you get here is just related to the map not being included in the test list, which we can fix once it's ready for merging. No other issues from Travis at the moment.

Zero-Bits:
This hasn't been touched for two weeks. Feel free to open it again if progress is resumed.

FaurnsGoldberg:
i'd like this to be easy to look at in case it should be resumed, even i might resume it.

PR #280, opened 2018-06-19
Removes Unused Away Sites

Zero-Bits:


PrPleGoo:
Bunch of deleted files in maps/away, nowhere else.
Assuming none of the deleted areas are used in saved maps; this should be fine.

PR #309, opened 2018-06-28
Reagent gas generation

NataKilar:
Heavily adapted port of https://github.com/Baystation12/Baystation12/pull/21703.

- The sublimator has been replaced with the gas generator, which uses chemical reagents instead of solid materials.
- A few new ice ores have been added as a temporary way to get new reagents such as nitrogen.
- The gas generator can combine reagents into gases (I.E. 2O + C = CO2). The current ratio is 10 reagent per gaseous mole
- Additionally, the gas generation has an aerosolization mode, creating a new gas out of chemical reagents.

This is reliant on https://github.com/Persistent-SS13/Persistent-Bay/pull/308, and this PR modifies some of the same files, but keeping it open for review on the concept, and also hopefully to help figure out how to save gas_data properly.


NataKilar:
This should be good, or at least ready for review. Still need input on gas_data saving, or initializing the reagent gases on server startup.

code/__defines/math_physics.dm
```diff
@@ -31,4 +31,4 @@
 #define TICKS_IN_SECOND 	10
 
 #define SIMPLE_SIGN(X) ((X) < 0 ? -1 : 1)
-#define SIGN(X)        ((X) ? SIMPLE_SIGN(X) : 0)
+#define SIGN(X)        ((X) ? SIMPLE_SIGN(X) : 0)
```
Zero-Bits: This file's changes seems unnecessary, considering they do nothing but remove the newline at the end of the file.

code/modules/materials/materials.dm
```diff
@@ -787,47 +777,54 @@ var/list/name_to_material
...
+	stack_type = /obj/item/stack/material/ice/nitroice
+	chem_products = list(
+				/datum/reagent/nitrogen = 20
+				)
+
+/material/ice/hydroice
+	name = "hydroice"
+	stack_type = /obj/item/stack/material/ice/hydroice
+	chem_products = list(
+				/datum/reagent/hydrogen = 20
```
Zero-Bits: Frozen hydrogen doesn't exist naturally; it's in a gaseous form even when in space. I'd mix this with some ice or water reagent. Or make some way to separate ice and/or water into hydrogen.

NataKilar: The idea where was it was hydrogen dissolved in water/ice. Adding ice and water reagents would've just made things more annoying, so I didn't think it was a good idea. Overall the whole ice->gas production is pretty unrealistic, and adding all the ice types is more of a temporary measure.

PR #325, opened 2018-07-02
Paper Additions

NataKilar:
- Paper Packages have been added to cargo, and contain 50 sheets to refill paperbins
- Paper shreds can now be placed into nanoprinters.


UndeadNappist:
Could we also get under /code/game/machinery/biogenerator.dm

		"Other" = list(
			/obj/item/weapon/paper = 5,
			/obj/item/weapon/paperpackage = 250,))

as well?

PR #328, opened 2018-07-10
Reagent Gas Improvements 

NataKilar:
General improvements to reagent gases.

- Reagent gases are now initialized on start-up, as with all other gases. No noticeable impact on performance
- Reagent gases (as a whole, not individually) can now be scrubbed and filtered in the respective machinery.
- Gases can now be converted back into reagents using the gas generator
- Only gases you are capable of making are now shown in the gas generator UI.
- Added a new cyanish canister type for reagent gases, selectable like all others.


NataKilar:
Almost certain the Travis errors are from upstream changes made in https://github.com/Persistent-SS13/Persistent-Bay/pull/327.

PR #330, opened 2018-07-21
Falsewall Airlocks

UndeadNappist:
Adds sprites and paths for airlocks that look like walls.

Re-introduces fake walls but as simply airlocks with a wall sprite, which is what they should have been in the first place anyway.

I probably broke something due to being exhausted. Whatever it is, I'll get to it in the morning.

image


UndeadNappist:
Runtimes are due to eyes.dm ; does not involve my own commit.

Is not yet a material_recipe barring further in game testing

PR #411, opened 2018-08-23
Stingy Lotto

rhicora:
Substantially reduce possible lotto payouts.

Choose from the following possibilities

`pick(150, 250, 500, 1000, 1500, 2000, 2500, 4000)`

Changed from:

`pick(5000, 10000, 50000, 100000, 500000, 1000000, 1500000)`


Andrew-Fall:
I believe there should be more lower picks, such as
pick(150, 250, 500, 1000, 1500, 2000, 2500, 4000)

rhicora:
Sounds fine to me.

PR #437, opened 2018-08-27
Healthier Spessmen

UndeadNappist:
Buffs the health of the Head, Chest, Lungs, Liver, and Heart to more realistic degrees.
It does not modify internal hemorrhaging or damage over time due to broken bones.
The head and chest are now EXTREMELY difficult to break or shatter, especially compared to the arms, feet, and hands.

In short, you will still take damage over time from internal injuries and broken bones, which will accumulate, and de-limbing but not dying is now easier.

### DIFF
--- a/code/modules/organs/external/head.dm
+++ b/code/modules/organs/external/head.dm
@@ -4,8 +4,8 @@
 	icon_name = "head"
 	name = "head"
 	slot_flags = SLOT_BELT
-	max_damage = 75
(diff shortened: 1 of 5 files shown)


UndeadNappist:
This gives you the opportunity to buff damage of glutslugs and other mobs, or make them target specific limbs.

For example, glutslugs could only target the legs or foot unless a target is downed, then go for the softer eyes.

PR #485, opened 2018-09-01
Freedom Ain't Free

Kurgis:
This PR effectively reverts all changes made in the infamous #243

I think the original points and reasons for the gun reversion has passed, it's time for the crew, of any faction out there, to cope with the existence of firearms. With the recent Mining changes, an undo of the plasma cutter debuff adding along with #243 also works well being removed. I know there may be some hesitance with the re-addition of AEGs, but I'm confidant a viable rework for it can be made before it becomes an endemic issue as in the last iteration.

https://i2.kym-cdn.com/photos/images/facebook/000/976/859/0bf.png


FaurnsGoldberg:
we're going to hold on this for a potential access/science rework. I'd like this to stay open so that i can merge it at that time.

Kurgis:
I'm going to wait until after the Science re-work to either re-open this PR, or create a new one.

PR #518, opened 2018-09-05
Make phoron experience more engaging

rhicora:
Adds the phoronation meter to mobs. Phoronated mobs have interesting lives.


UndeadNappist:
Please severely reduce or delete the phoron floating around in our sealed atmos container if this goes through.

PR #565, opened 2018-09-10
Changes researchable chameleon kit to full kit

idontthinksook:
Changes the researchable chameleon kit from the chameleon box to chameleon bag kit, doubles phoron prices and increases other material prices to compensate.

### DIFF
--- a/code/modules/research/designs.dm
+++ b/code/modules/research/designs.dm
@@ -864,8 +864,8 @@ other types of metals and chemistry for reagents).
 	desc = "A kit of dangerous, high-tech equipment with changeable looks."
 	id = "chameleon"
 	req_tech = list(TECH_ILLEGAL = 2)
-	materials = list(DEFAULT_WALL_MATERIAL = 500, "diamond" = 300, "gold" = 200, "phoron" = 10000)
-	build_path = /obj/item/weapon/storage/box/syndie_kit/chameleon
+	materials = list(DEFAULT_WALL_MATERIAL = 5000, "diamond" = 4000, "gold" = 4000, "phoron" = 20000)
+	build_path = /obj/item/weapon/storage/backpack/chameleon/sydie_kit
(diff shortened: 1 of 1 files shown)


Kurgis:
This would actually cause items contained in the Chameleon Kit to become unobtainable, as the backpack doesn't share them.

code/modules/research/designs.dm
```diff
@@ -864,8 +864,8 @@ other types of metals and chemistry for reagents).
 	desc = "A kit of dangerous, high-tech equipment with changeable looks."
 	id = "chameleon"
 	req_tech = list(TECH_ILLEGAL = 2)
-	materials = list(DEFAULT_WALL_MATERIAL = 500, "diamond" = 300, "gold" = 200, "phoron" = 10000)
-	build_path = /obj/item/weapon/storage/box/syndie_kit/chameleon
```
Kurgis: This would actually cause items contained in the Chameleon Kit to become unobtainable, as the backpack doesn't share them.

rhicora: Right. If we want to do this, the correct way to do it would be to edit the items spawned here:

https://github.com/Persistent-SS13/Persistent-Bay/blob/dev/code/game/objects/items/weapons/storage/uplink_kits.dm#L67

idontthinksook: The backpack should contain the Chameleon Kit itself, if you look directly above the kit in the file you looked.

rhicora: @Kurgis think this will be okay?

image

idontthinksook: So yeah, the backpack does share the items in the kit because it also contains the kit.

rhicora: There are also some questions about whether we *want* players to be able to freely wear any clothing item just because they have science access.

#code-discussion

PR #584, opened 2018-09-14
Time to hold a blood drive

rhicora:
_Vindictive_ change to encourage players to get each other donating blood.


Kurgis:
Blood loss is already a harsh mechanic, and I don't see how a change which is literally marked _**vindictive**_ is going to improve fun on any level from patient to doctor. This change also fails to mention the consequences of this change, and the realities that players must first face to even begin a simple blood drive. Characters, as far as I'm aware, all share different blood types which are incompatible with each other except for O-.

PR #592, opened 2018-09-15
Printer

UndeadNappist:
You can't build bioprinters anymore - hope you like harvesting organs!

The roborgan printer now uses osmium-carbide plasteel instead of normal steel for organs, so it's hideously expensive now.

### DIFF
--- a/code/game/machinery/bioprinter.dm
+++ b/code/game/machinery/bioprinter.dm
@@ -149,7 +149,7 @@
 	return O
 
 /obj/machinery/organ_printer/robot/attackby(var/obj/item/weapon/W, var/mob/user)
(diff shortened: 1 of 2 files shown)


ingles98:
Makes sense

SirBloodBuske:
I feel like you should keep the bio-printer and make it also expensive but that just my opinion

rhicora:
I think it's okay to remove the bioprinter completely, since there's still a roborgan printer for your expensive organ needs.

ingles98:
We'll let @FaurnsGoldberg decide.

UndeadNappist:
Hopefully that changed it to use OCP.

FaurnsGoldberg:
the bioprinter does other things than just organs. removing it without having a full rundown of how all its existing features are done elsewhere is not a good idea

PR #614, opened 2018-09-16
Add basic fire protection to engineering suits

Lance77am:


Kurgis:
I'm reluctant about a change which adds additional values to undersuits, but nonetheless I find this an interesting new feature.

UndeadNappist:
Phoron fires reach upwards of 25k K. Fires screw you up mostly bad by hurting all parts of your body at once - in this case, your hands, face, and feet are exposed.
The clothing are either flame-retardant (engineer jumpsuits at 800K is slightly below a typical fire, at 863K~ ish) with atmos jumpsuits having higher resistance.
The net result is that blasting an engineer in the arms, legs, and chest with a laser of a lit welder will do significantly reduced harm to them and, in the case of a fire catching them, will get away with lesser injuries so long as they don't stand around and let it boil their brain.

This is unlike actual FIRESUIT_MAX_HEAT_PROTECTION_TEMPERATURE which has a default K rating of 30000K, which is typical of both the atmos suit and and the fire proof suit.

I don't really view it as an issue.

I would also recommend a "catches fire and burst into flames!" proc for material/cloth, some other materials, and jumpsuits.

Kurgis:
I do not believe fire resistance should not affect energy or laser resistance, unless lasers also apply heat to the target (which would be applied separately from the damage anyways).

PR #623, opened 2018-09-17
Update girderlathe.dm

UndeadNappist:
Buffs girderlathe by reducing mat cost for girders from 4 to 2 to give people a good reason to use it.
Girders already give back 2 materials when deconned, so it shouldn't break the game.

To clarify:
Hand building takes 4 to construct a girder
Then 4 to reinforced.
Deconstructing the wall gives 2
Deconstructing the girder gives 2.

Using a girderlathe uses 2 material sheets instead of 4, meaning a total cost of 6.
Rgirders and the like still have to be adjusted by hand, but it's still cheaper to girderlathe -> reinforced girder part after placing.

### DIFF
--- a/code/modules/construction/girderlathe.dm
+++ b/code/modules/construction/girderlathe.dm
@@ -45,8 +45,8 @@
 					stored_materials[M.material.type] += M.amount
 					qdel(M)
 		for(var/M in stored_materials)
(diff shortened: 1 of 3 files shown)


UndeadNappist:
as per @EyesNoEyes request, added the ability to dismantle girders into girder parts.

PR #641, opened 2018-09-18
Frame Fixes/ Construction Fixes/ Label Increase/ALTCLICK Rotate

SirBloodBuske:
Fixes the bar sign frame, notice board, sinks. 
Increases label on pill bottle
adds alt rotate to rotatable objects
Adds piano crate
Comments out build able button Can't be used currently.
Changes world save text, to be seen.


rhicora:
I like this a lot.

code/datums/supplypacks/hospitality.dm
```diff
@@ -159,3 +159,11 @@
 	cost = 10
 	containertype = /obj/structure/closet/athletic_mixed
 	containername = "\improper athletics crate"
+
+
+
+/decl/hierarchy/supply_pack/hospitality/piano
```
ingles98: You did not mention on your details you were adding a new supply crate.

code/game/machinery/Sleeper.dm
```diff
@@ -235,10 +235,10 @@
 		occupant.client.perspective = MOB_PERSPECTIVE
 	occupant.dropInto(loc)
 	occupant = null
-	for(var/atom/movable/A in src) // In case an object was dropped inside or something
-		if(A == beaker)
+	for(var/obj/O in (contents - component_parts)) // In case an object was dropped inside or something. Excludes the beaker and component parts.
```
ingles98: You could've let it atom/movable. There might be a hidden reason as to why use only movable atoms (this includes some of the objects and all mobs)

code/game/machinery/pipe/construction.dm
```diff
@@ -319,6 +319,8 @@ Buildable meters
 		return ..()
 
 // rotate the pipe item clockwise
+/obj/item/pipe/AltClick()
```
ingles98: thank you for this hehe

code/game/machinery/wall_frames.dm
```diff
@@ -81,14 +81,15 @@ var/refund_type = /obj/item/stack/material/plastic
 	icon_state = "bulb-construct-item"
 	refund_amt = 1
 	build_machine_type = /obj/machinery/light_construct/small
-
+/*
 /obj/item/frame/driver_button
```
ingles98: The commenting/removal of this object wasn't clarified on the details. I do understand we might not even need it

code/game/objects/structures/noticeboard.dm
```diff
@@ -38,9 +39,12 @@
 			to_chat(user, "<span class='notice'>You pin the paper to the noticeboard.</span>")
 		else
 			to_chat(user, "<span class='notice'>You reach to pin your paper to the board but hesitate. You are certain your paper will not be seen among the many others already attached.</span>")
-	if(isCrowbar(O))
+	if(isWrench(O))
```
ingles98: Not sure if crowbar would make more sense for this one. Dropping the item is a nice addition

code/game/objects/structures/watercloset.dm
```diff
@@ -354,9 +354,16 @@
 	var/frame_type = /obj/item/frame/plastic/sink/
 	var/busy = 0 	//Something's being washed at the moment
 
-/obj/machinery/sink/New()
-	..()
+/obj/machinery/sink/New(loc, dir, atom/frame)
+	..(loc)
```
ingles98: i think it would be best to leave it param'less for this one. No use for whatever comes from the parent's "New()", the loc var is only handy for what you're adding next. I could probably be wrong

code/game/objects/structures/watercloset.dm
```diff
@@ -512,6 +519,17 @@
 	icon_state = "sink_alt"
 	frame_type = /obj/item/frame/plastic/kitchensink/
 
+/obj/structure/sink/kitchen/New(loc, dir, atom/frame)
```
ingles98: If its a direct copy of what you did to "sink", then you probably won't need it, since it literally calls "sink"s New(), sink/kitchen is sink's child therefore inherits its New() proc.

My eyes could be playing tricks on me but I see the pixel_x and pixel_y values are identical so this is mere redudancy, higher cpu usage (not that it would be noticeable) and possibly weird to find bugs in the future (plus weird maintaining)

code/game/objects/structures/windoor_assembly.dm
```diff
@@ -271,6 +271,9 @@ obj/structure/windoor_assembly/Destroy()
 
 
 //Rotates the windoor assembly clockwise
+/obj/structure/windoor_assembly/AltClick()
```
ingles98: Splendid

code/modules/economy/ATM.dm
```diff
@@ -148,6 +149,17 @@
 		to_chat(user, "\icon[src] <span class='warning'>The [src] beeps: \"[response]\"</span>")
 		return 1
 
+/obj/item/frame/atm/try_build(turf/on_wall)
```
ingles98: To all your try_build procs, i'm not sure if they work as you're not implementing on it, you're replacing it... Not sure if it would be better to check on them first for something that might be missing

code/game/machinery/wall_frames.dm
SirBloodBuske: Can't currently made or use.

PR #643, opened 2018-09-18
More holowarrant fixes, holowarrant crate

rhicora:
- Warrant Assistant will now check for either Security or Command access.
- Holowarrants can be swiped by a Security member or Command member (your station can decide who is actually authorized to sign off on warrants)
- Holowarrant crate is $500 and contains 4 warrant projectors.


FaurnsGoldberg:
are warrents and the wanted system factional? if so how

PR #648, opened 2018-09-18
Remotely revoke telepads

rhicora:
As a Supply Program user, if you have `core_access_command_programs` for the faction, you can use the Admin menu. The admin menu allows you to revoke the connection of rogue telepads.

Also fixes a bug where all possible import pads were selected if it didn't know which to use.


Andrew-Fall:
You can completely remove the IF statement

rhicora:
@Andrew-Fall Which one?

PR #663, opened 2018-09-20
merge to dev, exports buff

FaurnsGoldberg:


code/game/machinery/jukebox.dm
```diff
@@ -102,22 +102,23 @@ datum/track/New(var/title_name, var/audio, var/genre_name)
 /obj/machinery/media/jukebox/ui_interact(mob/user, ui_key = "jukebox", var/datum/nanoui/ui = null, var/force_open = 1)
```
rhicora: Need to make sure this is what is called instead of `tg_ui_interact`! Just remove the tg_ part of the thing that calls it, a few lines earlier

PR #667, opened 2018-09-20
Update new supplypacks.

JulianCarlosShirey:
Add two new supplypack:
-Bandage Pack Crate contain: 10 Bandages. 15 each. 150$ for the crate.
-Advance Trauma Kit Crate : 10 Trauma Kit. 30 each. 300$ for the crate.

Not sure about the objects name i a have just done some copy paste and some editing.
Tell me were i can find all the objects name please.


JulianCarlosShirey:
Is this the right way to to it?

ingles98:
The right way to do it is to test it mate.

PR #688, opened 2018-09-23
Smart fridge fix

PsyCommando:
Basically when you use a container on a smartfridge it may delete all incompatible items in the container. This fix just swaps the two checks so that they happen in the correct order and if the accept_check fails it won't just delete your things instead of just leaving them alone.


PsyCommando:
Okay reverted the turbine fix thing.. Sorry about that

PR #699, opened 2018-09-24
Factionalizes personal closets. Fixes Compile error

NataKilar:
Personal closets are now only unlockable by command staff _of the person's faction_. Fixes a compile error with a worth value for a nonexistent integrated circuit


rhicora:
This is good.

PR #705, opened 2018-09-24
Turrets should properly assess bots

rhicora:
Skreeee!


ingles98:
Try getting there the changelog file. Check my pr for an example

PR #712, opened 2018-09-24
Bay's Flags and Apollo's Phoron Desub

Andrew-Fall:
Port's Bay's massive flag rework along with Cakey's port of Apollo Phoron Desub directed towards Bay


code/modules/mob/living/carbon/human/examine.dm
```diff
@@ -270,23 +264,12 @@
 			perpname = name
 
 		if(perpname)
-			var/datum/world_faction/faction = get_faction(user.GetFaction())
```
SirBloodBuske: Not sure about faction stuff.

code/game/machinery/kitchen/microwave.dm
```diff
@@ -47,120 +48,77 @@
 		acceptable_items |= /obj/item/weapon/holder
 		acceptable_items |= /obj/item/weapon/reagent_containers/food/snacks/grown
 
-
```
SirBloodBuske: this is needed for building/decon

PR #713, opened 2018-09-25
Supplypack Update

Kurgis:
Restores names to ordered lockers, crates should no longer spawn blank. Ensured that all secure crates have an access requirement as so they do not gain null access.
All heads crates now require command access to open. The head of security order now requires leader access to open.
Added a new Bridge Secretary job order, giving access to command headsets.
Adjusts the pricing on various supply pack orders: cut down the price on livestock significantly, increased the cost of the jukebox, and added additional items to orders which were present in their containertype.
Removes guns from all operations orders. Removes the entertainment headsets from job orders, as they were designed for admin usage.


Kurgis:
Should be ready to merge after review.

PR #741, opened 2018-09-27
Fix passive gate limbo state

PsyCommando:
This should fix the issue where passive gates in space can't be used anymore after a server restart.
Keep in mind, I have no way to accurately test this, because the server version on the repo on either main or dev currently doesn't save properly at all out of the box. But, after reading it over several times it seems that's the issue.


ingles98 (changes requested):
The UI shouldnt be acessible if there is no power. You should look into a non hacky way of fixing the stat flag update. Id look into Process().

Add a bugfix changelog to your PR please.

PsyCommando:
Well the passive gate doesn't use power at all. Its unpowered.
But for some reasons, there was a check to see if its powered in ui_interact for it..

PsyCommando:
I double checked, and if the passive gate was even set to use power at all, it would drain 0 power, since the power draw vars are unchanged from the /obj/machinery/atmospherics base class.

ingles98:
I know they do not use power, what im saying is that if they cant be accessed after reboot, you should update the stat variable on after_load() as the flag could still be an issue somewhere else

PsyCommando:
But the flag is fine I think. Because in un-powered areas it really does have no power. 

The passive gate has the "use_power" var set to 0 the whole time which means it doesn't want power from APCs currently. And the NoPower flag is just whether or not the Area power/APC is supplying this machine afaik, not other forms of power like direct cable connection.
At least from reading the code for other atmos machines, that seems to work like that. Say solars and SMES don't care about the flag. 

```
/obj/machinery/power/smes/Process()
	if(stat & BROKEN)	return
	if(failure_timer)	// Disabled by gridcheck.
		failure_timer--
		return
[.....]
```
And
```
/obj/machinery/power/smes/ui_interact(mob/user, ui_key = "main", var/datum/nanoui/ui = null, var/force_open = 1)

	if(stat & BROKEN)
		return
```


The NOPOWER flag is never set manually in the passive gate code either.

Here's how the NOPOWER flag is computed too :
in power line 100:
```
// called whenever the power settings of the containing area change
// by default, check equipment channel & set flag can override if needed
/obj/machinery/proc/power_change()
	var/oldstat = stat

	if(powered(power_channel))
		stat &= ~NOPOWER
	else
		stat |= NOPOWER

	. = (stat != oldstat)
	if(.)
		update_icon()
```

NOPOWER will never be unset for unpowered areas. Its just made this way. So its not a hack to avoid the check if we're not even using power I guess? 
Unless you want me to add a check at the /obj/machinery level to see if the machine uses any power at all, and skip setting the flag in that case. But I don't know.

PsyCommando:
Thanks! And that sounds good!

PR #742, opened 2018-09-28
Constructable secure safes

Kurgis:
Adds secure safe frames, and adds them to the recipe list.


rhicora:
Testing of this is pretty important, make sure it retains its code between resets.

PR #748, opened 2018-09-28
Update windowdoor.dm

UndeadNappist:
Forgot to add the qdel(src) call, so it doesn't delete the old windoor when you deconstruct it. Oops.

### DIFF
--- a/code/game/machinery/doors/windowdoor.dm
+++ b/code/game/machinery/doors/windowdoor.dm
@@ -200,7 +200,7 @@
 			visible_message("<span class='warning'>The glass door was sliced open by [user]!</span>")
 		return 1
 	//If it's emagged, crowbar can pry electronics out. Now deconstructable!
-	if (src.p_open && isCrowbar(I))
+	if (isCrowbar(I) && p_open)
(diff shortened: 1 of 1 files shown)


ingles98:
Dupe exploit so this should get merged asap

code/game/machinery/doors/windowdoor.dm
```diff
@@ -217,7 +217,7 @@
 			wa.set_dir(src.dir)
 			wa.state = "02"
 			wa.update_icon()
-
+				qdel(src)
```
ingles98: This doesnt seem to be on the same level of identation. "STIGMA IS A WHINY CODER" because even the slightiest changes should be AT LEAST compile tested.

ingles98: If for some reason diff is telling me a different story, pardon me

SirBloodBuske: also the current way this is setup anyone can just take apart the windoor.

UndeadNappist:
Anyone can dismantle an airlock, too.

PR #761, opened 2018-09-30
Add duct tape crates

PsyCommando:
Added an orderable duct tape crate to the supply section in the cargo app.
Long overdue!


PsyCommando:
Can I get a Do not merge on this? I wanna make duct tape finite first.

ingles98:
> Can I get a Do not merge on this? I wanna make duct tape finite first.

Making duct tape finite should be on a different PR.

PsyCommando:
Yep, I'm making it on its own pr

code/datums/supplypacks/supply.dm
```diff
@@ -205,3 +205,9 @@
 	containertype = /obj/structure/closet/crate/secure/large
 	containername = "\improper Blueprints crate"
 	access = core_access_leader
+
+/decl/hierarchy/supply_pack/supply/duct_tape_crate
+	name = "Duct tape crate x10"
+	contains = list(/obj/item/weapon/tape_roll=10)
+	cost = 5
```
ingles98: The cost should be debated. Go on discord and ask what should be the price for each crate. Also, always start high expensive so that balancing issues are at least less exploitable

ingles98: @FaurnsGoldberg should give you an hint

PsyCommando: I already discussed it on discord the day I made this. But, I haven't heard brawler's opinion yet. And well, I tend to feel most things in those packs are incredibly expensive for what they are, which is why I went for that number. But if you guys wanna increase that to something higher I guess its fine. But making duct tape a rare luxury item might not be such a great idea, considering most of its uses are as a cheap substitute for something better.

PR #766, opened 2018-10-01
Finite duct tape

PsyCommando:
Tweaked duct tape to have a limited amount of uses.
Has 30 uses, but some usage will consume more than one use, like cuffing, gagging, blindfolding, crafting, repairing, and surgery using duct tape.  Sticking paper to walls and stuff still consumes a single use.

I wasn't sure if I should have used a datum or something for the numbers, considering how spread out duct tape uses were, so I just defined some macros with the needed quantities where it was used.


code/game/objects/items/weapons/tape.dm
```diff
@@ -1,14 +1,43 @@
...
 	desc = "A roll of sticky tape. Possibly for taping ducks... or was that ducts?"
 	icon = 'icons/obj/bureaucracy.dmi'
 	icon_state = "taperoll"
 	w_class = ITEM_SIZE_SMALL
+	var/uses_left = 30
+
+/obj/item/weapon/tape_roll/proc/has_enough_tape_left(var/amount_needed)
+	return (uses_left >= amount_needed)? TRUE : FALSE
+
+/obj/item/weapon/tape_roll/proc/delete_if_empty()
```
ingles98: Might as well not use a seperate proc for this small non repeated computation, at leat in my opinion since i only see it getting called on use, once.

code/game/objects/items/weapons/tape.dm
```diff
@@ -1,14 +1,43 @@
...
+
+//Meant to be used by external entities wanting tape for their purpose
+//It does a check to see if there's enough and then delete the roll if its empty after use
+//Return 0 when not enough tape, returns true when there was enough
+/obj/item/weapon/tape_roll/proc/use_tape(var/amount)
+	if(!has_enough_tape_left(amount))
+		return FALSE
+	uses_left -= amount
+	delete_if_empty()
+	return TRUE
```
ingles98: You might get into the awkward situation of this proc never returning true even tho it used the required amount, since its deletion got queued, in a perfect situation the object could get deleted immediately before the return TRUE. It may never ever happen but it is still not good practice expecting something to be called from a queued for destruction object

ingles98 (changes requested):
I will give you a better review later. Until then these are the only major issues. Overall this PR could very well go live as is. Code looks clean and the use of the macros was way more logical than a datum. Having the macros on the files that use them is better than setting up on a single file since we always know what we will be looking for on the files that matter. Since they are used on one file only, always #undef at the end.

ingles98:
Code reviewed. After you comment on it or make changes ill approve it.

code/game/objects/items/weapons/tape.dm
PsyCommando: So deleting an object also affects the call stack? That's a strange design choice.. But alright.

ingles98: Logically you shouldn't expect the call stack that resides on an object to keep on even after the object is deleted. It might not do any trouble as is, im not sure if the call stack stops immediately after its deletion but once it is queued you should at least not have it referencing the object obviously. In this case it is just a return, and perhaps the object will never instantly get deleted on the same tick as it is queued so no trouble

PsyCommando: Oh, wait, I see now. I'll just call qdel directly in the use_tape proc instead

code/game/objects/items/weapons/tape.dm
PsyCommando: Unless you mean delete_if_empty?

ingles98: Externally?

code/game/objects/items/weapons/tape.dm
ingles98: We also have the redudancy issue of asking if it has enough tape, and if it does, check again on use. I woud do something like changing the use_tape to stay as is and return 1 if it was successful, or any other positive int for the amount needed or simply 0 then "var/result = thetape.use_tape(AMOUNT)" and if the result is different than 1,its because it didn't work, then you could display it on a to chat saying how many more you need or how many you need (two different approaches). This is merely suggested, it is already functional as is

PsyCommando:
Okay, I applied the changes. I moved qdel into use_tape, so hopefully this should avoid issues with delete_if_empty possibly returning to possibly a now non-existent proc. And well, delete_if_empty was used only once.

ingles98:
> Okay, I applied the changes. I moved qdel into use_tape, so hopefully this should avoid issues with delete_if_empty possibly returning to possibly a now non-existent proc. And well, delete_if_empty was used only once.

Oh the "issue" is still present, its just for clearence. Nothing else should be checking if the tape reached 0, only after use

ingles98:
Confirm that you tested it thoroughly and it works as expected, and see if you wanna take my suggestion. Only then i'll be able to merge it to DEV

PsyCommando:
> 
> 
> > Okay, I applied the changes. I moved qdel into use_tape, so hopefully this should avoid issues with delete_if_empty possibly returning to possibly a now non-existent proc. And well, delete_if_empty was used only once.
> 
> Oh the "issue" is still present, its just for clearence. Nothing else should be checking if the tape reached 0, only after use

Hmm, not sure what you mean there? Is it fine for you or? I'll try to test it now.

PsyCommando:
Tested and it seems to work.

code/game/objects/items/weapons/tape.dm
PsyCommando: That's an interesting take. And I knew about the redundancy, but I didn't really see it worth making a fuss about since its literally a number comparison. But I can try your suggestion.

PsyCommando: I thought about it some more, and well, the issue I'd have if I'd do what you suggested is possibly ending up one day with a bug where the duct tape is consumed even if the action fails. If someone adds a new condition to the action, and the var/result = tape.use_tape(ammount) is pushed around or moved.

ingles98: If not enough tape return string "need extra [amount - uses_left]" (never uses) 
If enough return 1 (used amount)

ingles98: If result different than 1 to_chat result

PsyCommando:
Okay, so I changed a few places to use use_tape instead of doing another check for the amount of tape. And also re-tested everything, including surgery. Except I still kinda suck at it

ingles98 (approved):

PR #771, opened 2018-10-02
Revert "Update bioprinter.dm"

UndeadNappist:
This reverts commit a9aa3ccdf8a534eeeacd4032263cd1ed8db39495.

It is horrible to have the roboprinter nonfunctional. As such, I am doing a revert on this specific file. I will later get it to use OCP like I originally intended - but in my opinion, not having it available to players is worse at this time. Future fixes are to follow.

It has been fully tested and the roboprinter works without issues.

### DIFF
--- a/code/game/machinery/bioprinter.dm
+++ b/code/game/machinery/bioprinter.dm
@@ -133,8 +133,8 @@
 	stored_matter = max_stored_matter
 
 /obj/machinery/organ_printer/robot/dismantle()
-//	if(stored_matter >= matter_amount_per_sheet)
-	//	new /obj/item/stack/material/oc(get_turf(src), Floor(stored_matter/matter_amount_per_sheet))
+	if(stored_matter >= matter_amount_per_sheet)
+		new /obj/item/stack/material/steel(get_turf(src), Floor(stored_matter/matter_amount_per_sheet))
 	return ..()
 
 /obj/machinery/organ_printer/robot/New()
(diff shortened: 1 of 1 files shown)


code/game/machinery/bioprinter.dm
```diff
@@ -149,7 +149,7 @@
 	return O
 
 /obj/machinery/organ_printer/robot/attackby(var/obj/item/weapon/W, var/mob/user)
-	if(istype(W, /obj/item/stack/material/ocp) && W.get_material_name() == "matter_type")
+	if(istype(W, /obj/item/stack/material) && W.get_material_name() == matter_type)
```
ingles98: So it takes any type of material and dismantles strictly as steel?

UndeadNappist: Yes.

ingles98: So i can transmute wood into steel? Hehe in this economy it wouldnt be worth tho

ingles98:
What is the issue with OCP?

UndeadNappist:
It was no longer accepting any materials whatsoever after a series of changes. I'm currently looking into it, and an overall change to how organs rot, decay, and blood-type organ exchanges to make it more viable to harvest them from the recently (few hours) dead.

ingles98:
> It was no longer accepting any materials whatsoever after a series of changes. I'm currently looking into it, and an overall change to how organs rot, decay, and blood-type organ exchanges to make it more viable to harvest them from the recently (few hours) dead.

Well it was obviously because you had quoted that variable... Perhaps keeping it OCP and go with it ? If you didn't manualy revert this i'll tell ya what is the issue, you had "matter_type" instead of matter_type which prevented any material from being able to pass heh

UndeadNappist:
Tested it. Doesn't work. Leaving it /ocp disables the ability to insert any sheets into it whatsoever, including /material/ocp.
Still looking into it. In the meantime, this needs to be committed or else people literally cannot get replacement organs and all organ breakage will result in immediate death.

Issue #774, opened 2018-10-02
Items can be rotated when multiple tiles away


ourpar:
#### Description of issue
Items can be rotated by ALT + LMB click when your character more than one tile away

#### Difference between expected and actual behavior
You should only be able to rotate when standing next to the item.

#### Steps to reproduce 
1. find an item that can be rotated ( glass panes, mining braces, etc.)
2. Alt + Left click while more than one tile away
3. discover you have telekinesis.

#### Length of time in which bug has been known to occur

If I were to guess, the PR that added [ALT clicking](https://github.com/Persistent-SS13/Persistent-Bay/pull/641) is when it started (rip buske)

#### Issue bingo

Please check whatever applies. More checkboxes checked increase your chances of the issue being looked at sooner.

- [x] Issue could be reproduced at least once
- [ ] Issue could be reproduced by different players
- [x] Issue could be reproduced after a restart
- [x] Issue happened recently (less than 7 days ago)
- [x] [Couldn't find an existing issue about this](https://github.com/Persistent-SS13/Persistent-Bay/issues)

ingles98:
Hehe yes, i saw what was the cause right when I read the title. It is missing the distance check. Ill let newcommers work on that.

AlAtEX:
Shouldn't this issue be closed?

ourpar:
yep, thanks for the reminder.

PR #780, opened 2018-10-02
Fixed Rotation Range (And Rotating While Incapacitated)

AlAtEX:
Changed objects changed in the alt-click addition, not sure if more are affected, added 

if(!usr || !Adjacent(usr))
	return

if(usr.incapacitated())
        return

to said files under the rotate() function, should terminate the function if the player acting on it is not beside, or is incapacitated.

#774

### DIFF
--- a/code/game/machinery/pipe/construction.dm
+++ b/code/game/machinery/pipe/construction.dm
@@ -327,6 +327,12 @@ Buildable meters
 	set name = "Rotate Pipe"
 	set src in view(1)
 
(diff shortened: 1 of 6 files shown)


ingles98:
I will review this in a few hours. Thanks for your contribution and welcome!

ingles98:
Confirm you tested it thoroughly

AlAtEX:
Confirmed on testing for all six files, alt-click no longer functions like you have telekinesis, or functions while you are lying down/incapacitated.

PR #790, opened 2018-10-04
Properly identify borgos that don't have access

rhicora:
Currently, turrets will not shoot at any borgos if set to "check access". In this merge, check their access.


ingles98:
I don't get it, what does the title have to do with the change ?

rhicora:
I thought this was going to get merged right away by Brawler but it missed a deploy. I'll make it more descriptive.

PR #791, opened 2018-10-04
Allows mining borgs to traverse the asteroid

ourpar:
Engineering borgs have this var that allows them to traverse tiles without gravity, this should function exactly like that.


ingles98:
Tested ?

ourpar:
yep, I tested with: "/turf/simulated/asteroid" , "/turf/simulated/floor/asteroid" , "/turf/simulated/floor" and cyborgs with the mining module chip can move freely on the tested tiles

Issue #784, opened 2018-10-03
Microwave cannot be disassembled


PsyCommando:
#### Description of issue
Once constructed you can't disassemble the microwave at all. It just refuses the screwdriver as if it was a reagent.

Pretty sure its a simple fix, just need to check for the screwdriver and treat it differently.

#### Difference between expected and actual behavior
You should be able to unscrew the microwave and disassemble it.

#### Steps to reproduce 
use screwdriver on microwave

#### Length of time in which bug has been known to occur

More than a week.

#### Client version, Server revision & Game ID

Client Version: 512
Server Revision: be7cc741686a60c6f72fd166fea83bd8bb11de72 - main - 2018-10-02
Game ID: bWE-c3ea

#### Issue bingo

Please check whatever applies. More checkboxes checked increase your chances of the issue being looked at sooner.

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced after a restart
- [x] Issue happened recently (less than 7 days ago)
- [x] [Couldn't find an existing issue about this](https://github.com/Persistent-SS13/Persistent-Bay/issues)

PR #799, opened 2018-10-05
Dismantable microwave

PsyCommando:
Fixes issue  #784. The conditional had faulty logic. Honestly, the microwave class is a bit confusing and weirdly written, and I wish I could do more, but at least with those changes it can be taken apart now, like it always should have been possible to.


ingles98 (approved):
Spaget code gotta be fixed with more spaget heh.
Looks good to go for testing on DEV

ingles98 (changes requested):
Add a changelog entry specifying that people are able to deconstruct it.

PsyCommando:
Didn't feel it was needed since it literally should have always worked. The if statement was just ordered wrong.. And well, it didn't return in a few place it should have.

ingles98:
> Didn't feel it was needed since it literally should have always worked. The if statement was just ordered wrong.. And well, it didn't return in a few place it should ha
thing about it is that people are used to not have it working

ingles98 (approved):

PsyCommando:
> 
> 
> > Didn't feel it was needed since it literally should have always worked. The if statement was just ordered wrong.. And well, it didn't return in a few place it should ha
> 
> thing about it is that people are used to not have it working

True XD

PR #805, opened 2018-10-06
Auto Save Freeze, Cryo-Tube Fix 

SirBloodBuske:
Was having issues with my branch, However the PR is back!
image
image
image
image


Kurgis:
Does this also freeze ZAS, and mob processes during Save Freeze?

SirBloodBuske:
> Does this also freeze ZAS, and mob processes during Save Freeze?

From what I have tested yes.

ingles98:
This will need intense review on my part. Most of the code is working around the same logic i've seen and tested myself on your older PR on the subject. Overall it looks good and ill be commenting on what should be changed and whether its accepted.
Ill take your word that you have tested it thoroughly.

SirBloodBuske:
> This will need intense review on my part. Most of the code is working around the same logic i've seen and tested myself on your older PR on the subject. Overall it looks good and ill be commenting on what should be changed and whether its accepted.
> Ill take your word that you have tested it thoroughly.

Needs to be tested large scale, To see if the new system breaks anything.

code/game/machinery/cryo.dm
```diff
@@ -20,59 +20,33 @@
...
-	for(var/obj/machinery/atmospherics/target in get_step(src,node_connect))
-		if(target.initialize_directions & get_dir(target,src))
-			node = target
-			break
 
 /obj/machinery/atmospherics/unary/cryo_cell/Destroy()
 	var/turf/T = loc
 	T.contents += contents
-	var/obj/item/weapon/reagent_containers/glass/B = beaker
 	if(beaker)
```
ingles98: where does beaker exist?

ingles98:
look i do not understand why do you have a cryo change in this PR and it changes so much i cant even understand it

code/game/machinery/cryo.dm
SirBloodBuske: It exists at the top of the code

ingles98:
All these changes should be remade since the load/save mechanism as changed greatly. Please make sure it does not have anything else other than the Auto Save freeze featured changes, as i not only see cryo-tube stuff, it also has newspaper and xenobiological changes, it makes no sense. Overall the actual objective of this PR was met, it needs to be re-made upon the current mechanism code.

PR #810, opened 2018-10-07
Deconstructible bridge computer

PsyCommando:
Tiny fix for shuttle computers. It was missing its circuit entry in its definition. So the default base disassembly proc for computer would fail.


code/modules/shuttles/shuttle_console.dm
```diff
@@ -5,7 +5,7 @@
 	icon = 'icons/obj/computer.dmi'
 	icon_keyboard = "atmos_key"
 	icon_screen = "shuttle"
-	circuit = null
+	circuit = /obj/item/weapon/circuitboard/bridge_computer
```
ingles98: Hm im not sure about this... This should be a reference to the object itself, not a type, i could be wrong, but if that is true it shouldn't be that way.
I advise checking what the variable should actually contain. Is the variable defined as of that type or does it simply store the type ?
Also if it does store the type it should have been something more straightforward (not your wrong, just badly implemented by someone else in the first place) but i'm very sure that "circuit" should be holding the reference to the circuitboard, not the value of the type.

ingles98: **Ignore that.**
`var/circuit = null //The path to the circuit board type. If circuit==null, the computer can't be disassembled.`
It makes sense and actually describes why.
However making bridge computer deconstructable should be a balancing aspect that has to be discussed with @FaurnsGoldberg.
Sorry for wasting your time reading that heh :/

ingles98 (approved):

ingles98:
Needs @FaurnsGoldberg Approval.
If you already have his response, confirm it here please so i can merge it to dev myself.

ingles98 (changes requested):
I believe this one should also have a changelog entry

code/modules/shuttles/shuttle_console.dm
PsyCommando: Well, balancing wise. You can already cripple a shuttle very easily with regular engineer tools. Dismantle the engine, and rob the phoron in it, dismantle literally the entire shuttle but the console if you want. Either of those will cause significant balance issues already.
And well, you can already destroy the console too. So I don't see any reasons this would impact balance beyond having to rebuild a dismantled/removed console and reseting the info in it. This is mostly a convenience fix for shuttle builders that want to relocate their console without blowing it up, and wasting materials.

I think if there's balance work to be done, it would need to be done on the entirety of the shuttle system. And that's beyond the scope of this PR. This PR just falls in line with the current idea of allowing all shuttle parts to be freely disassembled.

ingles98 (approved):

PsyCommando:
The changelog was added by the way!

PR #811, opened 2018-10-07
Fix mirrored mixer and mirrored filters disconnecting on reload

PsyCommando:
Apparently the after_load() proc wasn't overriden for mirrored filters and mixers, so it would set the non-mirrored directions for the nodes to try and attach to other devices after load.


ingles98:
Was the issue actually just ONE of the directions being disconnected or would it create a runtime therefore not load any direction ? Nevertheless well spotted

ingles98:
Also did you test them ? Just need to confirm that hehe

PsyCommando:
Well, brawler asked me to pr the fix. I posted about it in the private code discussion on discord and showed what was happening.

And there's no way I can test it without the saved variables list. But list or not, its definitely an issue that the initial directions are changed on reload. Since the base class re-sets them, I'm assuming they're not saved. Or if they're saved, they used to get reset to the base class' values instead of the child class' values. Which is definitely not a good thing.

ingles98:
If the PR is ready confirm it and ill get it approved

ingles98 (approved):

ingles98:
this PR cannot be merged to Main without SoM on a partial dev-to-main merge. @FaurnsGoldberg 
it will produce nasty conflicts

PsyCommando:
Ah crap, I knew I probably should have based it off main..

PR #813, opened 2018-10-07
Fix helmet + magboot duping on future spacesuits

ourpar:
Eyes mentioned that /prepared suits dupe their helmet on restarts and its not something that can be fixed with saving vars, so this changes all voidsuits orders to contain helmet+suit+magboots instead of /prepared suits

### DIFF
--- a/code/datums/supplypacks/atmospherics.dm
+++ b/code/datums/supplypacks/atmospherics.dm
@@ -106,7 +106,9 @@
 //void suits
 /decl/hierarchy/supply_pack/atmospherics/voidsuit
 	name = "Voidsuits - Atmospherics Voidsuit"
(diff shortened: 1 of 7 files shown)


ingles98:
Indeed this is straightforward and i dont find it too hacky of a fix. I do believe prepared suits arent working properly and this will reduce the hassle while it doesnt get fixed.
Can you just confirm that you tested every each crate ? 
The types so far look right, but unless its confirmed that the crates properly contain those items i cannot quickly merge it to dev and perhaps even to main.
Can't test it myself at the moment and i currently have tons of work to do so depending on myself testing it would take long.

ourpar:
it looks like kurgis's cargo refactor (#815) does this and then alot more, so unless that PR doesn't get merged this one is pretty obsolete (I'll close this pr whenever that one gets merged)

PR #814, opened 2018-10-07
Higher jam chance for ballistics, lower miss chance for lasers

rhicora:
Balances lasers and ballistics while still giving each its own advantages. Guns hit harder, but are prone to jam - particularly when fired in burst mode. Lasers are pretty accurate, don't jam, and you don't need to manufacture their ammo, but they are weaker.

- 5% of firing attempts will cause a jammed weapon, which must be cleared by reloading.
- For automatics, a much higher chance, particularly when on burst mode.
- Lasers are now much more accurate at long range.

To clear a jam, you need to unload the current magazine and wait 4 seconds. If in the future we have a skills system, people with a "shooting" skill will have a lower jam chance.


ingles98:
~I cannot approve it myself since these are balancing changes, and quite impactful.
Only @FaurnsGoldberg will be able to approve this at all, be it for testing or Main.

rhicora:
@ingles98 Changelogs added for this and 3 other pulls

PR #815, opened 2018-10-07
Cargo refactor

Kurgis:
- Created new cargo tabs, along with various new orderables. This being flooring, custodial, livecargo, non-essentials, and assignments.
- You can now order ten copper through the materials cargo tab.
- Removed the miscellaneous cargo tab.
- Removed empty regeant cartridge order as you can craft them.
- Removed various \improper tags and labels from cargo which didn't need them.
- Removed mention of the word crate or order from cargo orders, as it's already implied they will come in such a container.
- Adjusts the price of job crates, after moving their gear into other orders.
- Re-arranged cargo contents into various tabs, operations now focuses on the actual operation of Cargo. - Job equipment/clothes have been moved into the new assignments tab. Various orders split between non-essentials and hospitality, with hospitality focusing on the usual duties of a Kitchen and Bar.
- Prepared voidsuits will no longer come in supplypacks, as they cause issues when they are altered. Likely the first step of their total removal.
- You may now order a fax machine through cargo, usage result may vary.

Props to Baystation12/Baystation12#22904 which I based some deal of the changes on.

### DIFF
--- a/code/datums/supplypacks/dispcarts.dm
+++ b/code/datums/supplypacks/dispcarts.dm
@@ -2,8 +2,7 @@
 	name = "Dispenser Cartridges"
 
 //refill packs
-
-/decl/hierarchy/supply_pack/reagents/alcohol_reagents
+/decl/hierarchy/supply_pack/dispenser_cartridges/alcohol_reagents
 	name = "Refills - Bar alcoholic dispenser refill"
 	contains = list(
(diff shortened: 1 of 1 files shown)


Kurgis:
After a discussion with Brawler, the dispenser cartridges were properly adjusted as to not put Bartenders out of business. The price of individual dispenser cartridges are now quite proportional to the price of their respective refill crate, and certain medicines have been placed back to their original value of 15.

PR #816, opened 2018-10-07
Do it fast, do it right. Surgeons: choose one.

rhicora:
Allows surgeons to be slightly faster, but incur a fail chance. Slows down surgery when doing things by the book. By default, surgical tools are set to **Careful** mode; click on them in hand to toggle to **Rushed** mode.

If skills are added, this is a good future opportunity to encourage specialization; vary the multiplier for time doing careful (or failure chance for rushing) based on surgery skill.


rhicora:
WIP, hang on a sec.

html/changelogs/rhicora-pr-816.yml
```diff
@@ -0,0 +1,33 @@
...
+author: rhicora
+ # Optional: Remove this file after generating master changelog.  Useful for PR changelogs that won't get used again.
+delete-after: True
+ # Any changes you've made.  See valid prefix list above.
+# INDENT WITH TWO SPACES.  NOT TABS.  SPACES.
+# SCREW THIS UP AND IT WON'T WORK.
+# Also, all entries are changed into a single [] after a master changelog generation. Just remove the brackets when you add new entries.
+# Please surround your changes in  double quotes ("), as certain characters otherwise screws up compiling. The quotes will not show up in the changelog.
+changes:
+  - tweak: "Added rushed and careful mode to most surgical tools (click in hand to toggle)"
```
ingles98: While the PR is approved and not merged, i'd suggest pointing out the fact that the steps will be taking twice as much time than before, and that rushed will be even faster than the vanilla times.

ingles98:
I believe it makes sense having this categorized as a **New Feature** instead of a Fix.
I will review this more thoroughly again now.

Kurgis:
Does the tool take into account when it leaves a user's hand, or does it remain in the usage mode even when dropped/set down?

FaurnsGoldberg:
Added in another PR

PR #820, opened 2018-10-08
Do not distribute Persistence 13 hub text

rhicora:
This should be in a config file that the genuine PS13 owner maintains.

Do not merge this until @FaurnsGoldberg creates the config entries.


rhicora:
When someone grabs the codebase and starts an unofficial server, they automatically hub with the following:

image

rhicora:
This is incredibly confusing to players who want the official server, @EyesNoEyes described it as deceptive though I'm sure it's not intentional. We just need to stop distributing the hub description.

It should be in our own individual config.

ingles98:
alright true

PR #826, opened 2018-10-09
copper 10x crate.

HbsRed:
Adds copper crate 10x at the price of silver. Copper is a commodity on the frontier as a valuable building resource and it is sold as such. Will keep the load off of buying the electrical maintenance crate /just/ for the wires it gives which also comes with the super load of clutter (power cells, gloves, flashlights, etc..)


ingles98:
These changes are conflicting with the current branch. You need to apply them at the top of the most updated dev branch.
It might also conflict with current supply pack PRs that are kn queue of being merged.

PR #831, opened 2018-10-09
Re fix pressure gates limbo

PsyCommando:
This time I tested it again and again almost the entire afternoon!
Ui shows, gate works too when in an area with no power!


ingles98:
The other one is already merged to dev right

PR #878, opened 2018-10-14
Mechfix

PsyCommando:
So with the help of @ingles98 I was able to fix all issues that mechs seemed to have on save reloading.
Also removed the horrible hoverpod noise when it moves.. Everyone call me out over ear rape when I move near them, and I don't blame them.
Also fixed a runtime in the portable_connector, since it was accessing the connected entity's internal variable instead of using a proc made just for that..

Here are the saved variables I'd need added for these to work properly (AKA I tested with those)

For /obj/mecha:
```
internal_tank
contents
equipment
cell
dna
health
occupant
internal_damage
cabin_air
use_internal_tank
lights
dropped_items
state
maint_access
connected_port
```
All those ensure the mech, whatever state it may be in will be reloaded in the exact same state.

For /obj/effect/decal/mecha_wreckage:
```
salvage_num
crowbar_salvage
wirecutters_salvage
welder_salvage
```
Those are mainly to keep people from harvesting new parts from wreckages on each reload.


ourpar:
mech fix sounds cool, but wouldn't having hoverpods completely silent be kind of OP in terms of stealth? even stepping barefoot makes some sort of noise

PsyCommando:
> 
> 
> mech fix sounds cool, but wouldn't having hoverpods completely silent be kind of OP in terms of stealth? even stepping barefoot makes some sort of noise

Well jetpacks don't make noise. But if you can find a sound that doesn't blow everyone's eardrums out, I could use it instead.

Andrew-Fall (approved):
Everything looks fine, however I believe approval should be received from Brawler as well before merging.

PsyCommando:
Brawler seems to approve!

PR #894, opened 2018-10-16
Adjusts Guns

### LINKED ISSUE #834: it is not possible to remove the temperature gun's device cell
#### Description of issue
whenever trying to remove the power cell from a temperature gun, it instead opens the UI to change the temperature setting rather than removing the cell

if the cell housing is open, it should remove the device cell rather than opening the gun's UI

#### Steps to reproduce 
1. obtain temperature gun
2. (optional) deplete gun's charge
3. open cell housing with a screwdriver
4. attempt to interact with the gun to remove the gun cell, but instead be greeted with the UI to change temperature settings.

temperature gun

i just discovered it today, however it's seems likely the update that introduced removable cells caused the issue.

Client Version: 512
Server Revision: 9c0f2baf0d7ac01ce6e79433e89410acce257dd3 - main - 2018-10-09
Game ID: bWL-aIcg

#### Issue bingo

- [x] Issue could be reproduced at least once
- [] Issue could be reproduced by different players
- [x] Issue could be reproduced after a restart
- [x] Issue happened recently (less than 7 days ago)
- [x] [Couldn't find an existing issue about this](https://github.com/Persistent-SS13/Persistent-Bay/issues)

### LINKED ISSUE #842: plasma cutters have removable power cells
#### Description of issue
plasma cutters, which are supposed to be fueled with phoron, have removable power cells, which allows you to recharge them with power rather than phoron

#### Steps to reproduce 
1. obtain plasma cutter
2. (optional) shoot it to lose charge
3. remove powercell like any other energy weapon
4. realize phoron powered plasma cutters do not actually need phoron to be powered

plasma cutter

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced after a restart
- [x] Issue happened recently (less than 7 days ago)
- [x] [Couldn't find an existing issue about this](https://github.com/Persistent-SS13/Persistent-Bay/issues)

NataKilar:
Energy weapons can no longer have their cells removed from their housing. Recharge them in the newly buildable recharger instead!

Fixes #842 
Fixes #834 
Fixes a bug in which cell rechargers would delete their cell upon deconstruction
Fixes a bug in which guns (both energy and ballistics) would have their loaded magazine or cell reset on load
Fixes a bug in which magazines and bullets would reset on load


UndeadNappist:
It's nice that you're adding rechargers, but I don't see the point in making energy cells non-removable, especially given the comparison to the current aspects of ballistics.

You're tackling a phantom problem that doesn't exist and probably won't, since the real issue is lack of prone mechanics and the overall high power of ballistics.

PsyCommando:
The bugfixes are fine, but I don't think we should make cells un-removable again. Its such an issue for energy weapons to need to be recharged instead of reloading them. It makes then unreliable in a pinch when there are many other ballistic alternative that offer a way to quickly and painlessly be reloaded. You don't want to be the one to run out of ammo in a firefight.

NataKilar:
A few things to note about this change. This is not intended to adjust the overall balance of energy weaponry,  and will likely leave them in the same state or slightly worse. The rationale behind this change is that the ability to remove power cells was intended solely as a method to recharge energy guns _at all_, and was not intended to make an "ammunition" like system for energy weapons.

If energy weapons are made re-loadable (they were not easily so previously, but others have stated their intent to generalize the energy gun cells) they become far more similar to ballistics. The end result will be that one type will always be better than the other, rather than fulfilling different niches. Thus, I believe that making this change will realign the purpose of energy weaponry in order to facilitate future balance changes. (If requested, I can move the cell removal stuff to a separate PR)

NataKilar:
Because some of these bug fixes are closely aligned with the removal of replaceable energy cells, I left it all in one PR. Since energy weapons are currently weak enough that they are barely used, and the "re-loadable lasers strategy" is probably used even less, I don't think this will have a major effect on balance.

code/game/machinery/recharger.dm
```diff
@@ -11,18 +11,43 @@ obj/machinery/recharger
...
-		if (istype(G, allowed_type)) allowed = 1
+
+	checking_allowed:
+		for(var/allowed_type in allowed_devices)
+			if(istype(G, allowed_type))
+				for(var/disallowed_type in disallowed_devices)
+					if(istype(G, disallowed_type))
+						break checking_allowed
+				allowed = 1
+				break checking_allowed
```
Andrew-Fall: This is seems overly complex, perhaps just check if it's disallowed first and return, then check if it's allowed.

code/modules/projectiles/ammunition.dm
```diff
@@ -17,8 +17,9 @@
 
 /obj/item/ammo_casing/New()
 	..()
-	if(ispath(projectile_type))
-		BB = new projectile_type(src)
+	if(!map_storage_loaded)
+		if(ispath(projectile_type))
+			BB = new projectile_type(src)
```
Andrew-Fall: I am fairly certain map_storage_loaded will only work in Initialize(), not New()

code/modules/projectiles/guns/projectile.dm
```diff
@@ -39,12 +39,13 @@
...
-				loaded += new ammo_type(src)
-		if(ispath(magazine_type) && (load_method & MAGAZINE))
-			ammo_magazine = new magazine_type(src)
+	if(!map_storage_loaded)
+		if (starts_loaded)
+			if(ispath(ammo_type) && (load_method & (SINGLE_CASING|SPEEDLOADER)))
+				for(var/i in 1 to max_shells)
+					loaded += new ammo_type(src)
+			if(ispath(magazine_type) && (load_method & MAGAZINE))
+				ammo_magazine = new magazine_type(src)
```
Andrew-Fall: Same as above, map_storage_loaded won't function in New(), only Initialize()

Andrew-Fall (changes requested):

NataKilar:
Closing this, going to split it into two PRs

PR #914, opened 2018-10-18
Readds hand-rolled cigarettes crafting recipe and other smoking related items

tacct6:
* Hand-rolled cigarettes are once again obtainable with cardboard
* "corn cob smoking pipe" was added as a crafting recipe, requiring wood, 
* Generic "smoking pipe" was added to "Clothing - Luxury accessories" because it looks very luxurious
* Slightly altered the "smoking pipe" mob sprite because it looked too close to the mouth when facing east and west


Andrew-Fall (approved):
Code checks out and travis passes.

PR #918, opened 2018-10-18
Fixes various bugs with guns

### LINKED ISSUE #834: it is not possible to remove the temperature gun's device cell

### LINKED ISSUE #842: plasma cutters have removable power cells

NataKilar:
Fixes #842
Fixes #834
Fixes a bug in which cell rechargers would delete their cell upon deconstruction
Fixes a bug in which guns (both energy and ballistics) would have their loaded magazine or cell reset on load
Fixes a bug in which magazines and bullets would reset on load


Andrew-Fall:
Everything checks out, however a changelog is required.

code/modules/projectiles/guns/energy/special.dm
```diff
@@ -207,6 +207,10 @@ obj/item/weapon/gun/energy/staff/focus
 
 
 /obj/item/weapon/gun/energy/plasmacutter/attackby(obj/item/A, mob/user)
+	if(isScrewdriver(A))
+		to_chat(user, "<span class='warning'>The cell housing is firmly secured, you can't remove it.</span>")
+		return
+
```
Andrew-Fall: It looks as though you forgot to split this change

NataKilar: This change is needed to stop plasma cutters from being able to have removeable power cells, as phoron is used to charge them instead. If the second PR goes through, then this change would be defunct.

PR #921, opened 2018-10-19
Craftablesv1

HbsRed:
Expect more of these craftables pr's


Andrew-Fall:
Checks out from a code perspective, but you'll need to ask Brawler if these changes are allowed. They appear to make everything extremely easy and cheap to get, with the exception of defibrillators.

HbsRed:
No theyre supposed to be makeable. I think i fixed it

PR #925, opened 2018-10-20
Add extra control to vents via air alarms

PsyCommando:
I added a way to set the internal pressure limit on individual vents via the air alarm. Up until now, you could toggle the internal pressure check on all air alarms for vents, but you couldn't actually set the actual internal pressure limit for some weird reasons..
So I added a way to set it via the usual air alarm ui.

Also added a way to toggle off breach detection and danger levels detection, so anyone can setup an alarm strictly for monitoring and etc..

I recommend adding the following saved vars:

For /obj/machinery/atmospherics/unary/vent_pump:
```
pressure_checks
internal_pressure_bound
```
For /obj/machinery/alarm:
```
report_danger_level
breach_detection
```


PsyCommando:
I wouldn't say its a new feature. Vents and alarms haven't really changed much at all. Its just the UI that wasn't displaying some of the stats and commands it should have.

michaelchessall (approved):

PR #927, opened 2018-10-20
Fix passenger enter verb on mech

PsyCommando:
So after a mech with a passenger compartment reloads, the "enter passenger" verb is never set again so passengers can't climb in unless you re-install the compartment. This fix does the check to add the verb after loading the save so hopefully that fixes the issue.


Andrew-Fall (approved):
Code checks out and Travis passes.

PR #928, opened 2018-10-20
Designer Update

ingles98:
Done.

Improves canvases by enabling us to edit them. **(Also titling and authoring added)**

Also making preparations for code refactor which will enable for faster process and bigger icon sizes.

### DIFF
--- a/code/modules/designer/designer.dm
+++ b/code/modules/designer/designer.dm
@@ -21,7 +21,7 @@
 	set category = "Debug"
 	design.pixel_width = text2num( input("Max Width:", "Designer DEBUG", 32) )
 	design.pixel_height = text2num( input("Max Width:", "Designer DEBUG", 32) )
-	design.Design()
+	design.Design(icon_custom)
 
 /obj/item/design_item/GetDesign(var/icon/ico, ckey)
 	if (!ico || !istype(ico) )
@@ -30,7 +30,8 @@
 	var/offset_y = text2num( input("Offset Y:", "Designer DEBUG", 0) )
 	ico.Shift(EAST, offset_x)
 	ico.Shift(SOUTH, offset_y)
-	icon = ico
+	icon_custom = ico
+	icon = icon_custom
 
 	if (ckey)
 		designer_creator_ckey = ckey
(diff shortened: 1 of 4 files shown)


code/modules/designer/designer.dm
```diff
@@ -45,18 +46,53 @@
...
 	if (source)
 		my_atom = source
 
 
-/datum/designer/proc/Design()
+/datum/designer/proc/Design(var/icon/icon_custom = null, offset_x = 0, offset_y = 0)
 	usr << browse(file("html/designer.html"), "window=designer;size=600x400;can_close=1" )
-	spawn(10)usr << output("href='?src=\ref[src];[pixel_width];[pixel_height]","designer.browser:getData")
+	sleep(5)
+	usr << output("href='?src=\ref[src];[pixel_width];[pixel_height]","designer.browser:getData")
```
Andrew-Fall, line 59: to_browser(user, output(...))

code/modules/designer/designer.dm
```diff
@@ -45,18 +46,53 @@
...
+	if (length(title) >= 1)
+		usr << output("[title]","designer.browser:setTitle")
```
Andrew-Fall, line 61: to_browser(user, output(...))

code/modules/designer/designer.dm
```diff
@@ -45,18 +46,53 @@
...
+		icon_custom.Shift(EAST, -offset_x)
+		icon_custom.Shift(SOUTH, -offset_y)
+		pixel_list = new/list(icon_custom.Width(),icon_custom.Height())
+		for ( var/x = 1 to icon_custom.Width() )
+			for ( var/y = 1 to icon_custom.Height() )
+				var/list/color = ReadRGB(icon_custom.GetPixel(x,y))
+				if (!color || !color.len ||color.len < 3)
+					continue
+				pixel_list[x][y] = color
+		spawn(2) usr << output("[json_encode(pixel_list)]","designer.browser:getIconAsJSON")
```
Andrew-Fall, line 74: to_browser(user, output(...)) Also why is spawn needed for these?

Andrew-Fall (changes requested):

code/modules/designer/designer.dm
ingles98: Simply the fact that it takes long in between the JS functions. Earlier versions i had problems with it as the order would get scrambled as some functions took more time than others, as they are called asyn it seems.
Havent tested with the newer approach. 

Also the multiple outputs are for multiple function calls with different arguments. I could indeed attempt a better implementation now as the code greatly diverged from what it was, but i only want to bother with it when i get the JSON string to be directly sent to DM

EDIT:
Oh yes i didnt know about the to_browser ss13 proc. Does it grant any benefit other than being friendlier?

Andrew-Fall: Not really, it's pretty much for safety reasons. Keeps travis happy

ingles98: Dunno if its only on mobile but searching for 'To_browser' finds absolutely no reference

PsyCommando: Can confirm, to_browser doesn't exist @Andrew-Fall

ingles98: I still haven't understood as to why Travis is mad about the PR.
"764 << uses (expecting exactly 762)" is the only thing I see pointing towards something off in code quality yet it doesn't tell me anything.

PR #933, opened 2018-10-21
Gives suit cyclers core accesses (and some other fixes)

tacct6:
1. New suit cyclers now use core accesses instead of old, nonassignable accesses

2. Air tanks now spawn empty when printed in the autolathe, no more free air. (the proc that was being used seemed like a non-functioning edited copypasta from printing empty cells in research designs)

3. Printing 1x cable from the autolathe no longer prints an entire stack, this prevents duping copper when recycling cables back into the autolathe

4. Prox sensors no longer need "waste" material to be printed in the autolathe


PsyCommando (commented):
Looks good to me! Thanks for fixing it! I didn't even get around to it yet..

michaelchessall (approved):

PR #942, opened 2018-10-21
Fix accessory verbs not restoring after save

PsyCommando:
Should fix holsters and various other accessories losing their verbs on save reload.


PsyCommando:
I think I found an issue...
Let me run some more tests before merging

michaelchessall (approved):

Issue #940, opened 2018-10-21
Free Air Tanks


tinfoiltophat:
#### Description of issue
When building air tanks with the autolathe, they do not require resources.

#### Difference between expected and actual behavior
One would expect air tanks to have some sort of cost in metal, but they do not.

#### Steps to reproduce 
1. Open autolathe menu.
2. Scroll to Air Tank
3. Build to your heart's content.

#### Specific information for locating

says No Resources Required next to the entry on the lathe

#### Length of time in which bug has been known to occur

N/A

#### Client version, Server revision & Game ID

Client Version: 512
Server Revision: b485ef2cf9b8dcd5e1cfabb1c915e1e412cf467f - main - 2018-10-21
Game ID: bWW-cF4B

#### Issue bingo

Please check whatever applies. More checkboxes checked increase your chances of the issue being looked at sooner.

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [ ] Issue could be reproduced after a restart
- [x] Issue happened recently (less than 7 days ago)
- [x] [Couldn't find an existing issue about this](https://github.com/Persistent-SS13/Persistent-Bay/issues)

astralenigma:
Weird, it should auto-close when the commit is made.

PR #943, opened 2018-10-22
Fix: Airtank now costs 250 steel

eternia16:
### Changes:
* Air Tank now costs 250 steel instead of nothing. Issue #940  (So they're no longer able to be created from thin air haha)

Really simple commit, nothing special.


Andrew-Fall:
I believe 250 steel is only 1/4th a steel sheet (1000 per)

eternia16:
Could be changed to a higher value if you want

JulianCarlosShirey:
emergeny air tank cost 280 steel so i would say atleast 500 for the air tank or more. you can store 4 time more oxy so making it even more expensive will not pose a problem. maybe around 800 steel?

eternia16:
Sounds fair to me, I've upped the requirements to 850 steel.

PsyCommando:
How did you manage to crash travis? ^^;
Did you use spaces instead of tabs? And did you try compiling it on your end?

eternia16:
I just updated the branch from the DEV, the checks passed before I did that haha.

eternia16:
https://imgur.com/a/BFlOCDE

PR #944, opened 2018-10-23
Welding torches now contain 0 fuel when printed from the autolathe + other welding changes

tacct6:
**Changes:**

- Welding tools printed in the autolathe no longer inlude fuel

- Added miniature welding tool to the autolathe, it only can carry 5 units of fuel but can be placed in storage containers that standard welding tools wouldn't fit in.

- Added huge welding tool to autolathe that can hold 80 units of fuel, the downside is it can't be placed in backpacks or your belt slot.

- Custom/painted welding masks + backpack welding tank are added to cargo imports, under engineering tab, because who says you can't be stylish on the job?!

- Adds the ability to insert reagents into the protolathe and fixes decloner gun recipe (not even sure if it's possible to get such high tech levels legitimately, but might as well fix it incase it is)

- Increases experimental welding tool production cost, item size, and slightly lowers its self-refuel rate.
         _I tested the reagent requirement for printing it and it works fine, in fact it works so good it makes me want to add reagent requirements for other items!_ 
----------------
IMO the experimental welding tool shouldn't be a thing because of the self regenerating fuel, but a fair amount of people could get upset that random pr 12408 removed "x" item from the game without much discussion or clear approval, so i figured the increased size and higher production costs is a fair tradeoff for the time being. (and a pretty fun tradeoff too, more protolathe recipes should use reagents!)


ourpar:
travis doesn't like my code :v 
i'll fix it

ourpar:
Ok, travis is happy and I tested all the welding starting amounts when printed from a lathe + refueled them without any issues, so this is ready for review

ourpar:
Ok, travis is happy, I'm happy, this is ready for review

ingles98:
Makes sense

michaelchessall (approved):

PR #953, opened 2018-10-24
Makes internals crate less useless plus huge scrubber

PsyCommando:
Changed the content of internal crates from gas masks + half filled AIR tanks(Which have to have a distro pressure of 101kpa to be of any use) to regular breath masks with double emergency oxy tanks, like on most ss13 servers.
To compensate, I made a crate of only assorted gas masks so they're still obtainable in bulk and you get to chose between half-mask and full gas masks.

Also made a crate for the huge scrubbers, which are meant to be used by an air control computer in atmos usually. But I'm sure people will loooove those, especially our friends at Cythos!


michaelchessall (approved):
looks good

PR #954, opened 2018-10-24
Fix cargo train power cell and key

PsyCommando:
Makes the power cell and keys in the cargo train not respawn on map load.
Be sure to add the key variable to the saved var list, along with the base class's cell variable for this to work properly.


michaelchessall (approved):
looks good

PR #955, opened 2018-10-24
Prevent microwave from spilling its gut when disposing content

PsyCommando:
So before when you'd press dispose content in the microwave menu, it would sometimes just spill out micro lasers and microwave boards and stuff, because that's stored in the "content" var along with the ingredients. So this uses a function in the machinery base class that's less used than it should "InsertedContents()" to only eject ingredients and not parts, ever.


michaelchessall (approved):
looks good

PR #956, opened 2018-10-24
Raise Temperature for Unathi Heat Comfort Spam

Koollan:
Raises the temperature trigger to spam comfort messages for Unathi from 295 to 305, as Unathi are almost always at 295 temperature or more.


Draxtheros:
For reference, room temperature is 293, and clothing drags you above 295 easily, spamming you with messages.

PR #959, opened 2018-10-25
Add co2 and chlorine to the gas filtered by gas masks

PsyCommando:
I thought it was kinda weird that co2 wasn't filtered when its been one of the major threats right now on station. And well chlorine gas was one of the gases used in warfare, and also to be countered with gas masks, so makes sense to me I guess.


PsyCommando:
Welp, I didn't realize, but I forgot to actually add a commit to the branch ^^;;

PR #970, opened 2018-10-26
Add co2 and chlorine to the gas filtered by gas masks REEEEEEEdux

PsyCommando:
Last pr I made I forgot to actually commit the files..


michaelchessall (approved):
looks good and actually makes changes

PR #992, opened 2018-10-30
Adds the advanced flash.

Pandolphina:
:cl: Pandolphina
   rscadd: Added the advanced flash to the research designs.
   /:cl:

Hopefully should encourage more use of the flash, and less "instant shotgun use" by security, or mainly CEOs.


Pandolphina:
Okey TortoiseGit is still borken and the web editor doesn't work, so I'll add this eventually.

PR #993, opened 2018-10-30
Fix cells and air content respawning on reload

PsyCommando:
* Fix cells and air content respawning on reload in portable scrubbers and pumps.
* Added descriptions to both and the Huge scrubber.

Cells and air content respawned on reload in the portable scrubbers and pumps. Changed that so it doesn't happen anymore. Also added  helpful descriptions for the pump, scrubbers and huge scrubbers.


Andrew-Fall:
Runtimes during load
`[00:00:00] Runtime in portable_atmospherics.dm,50: Cannot read null.volume
  proc name: MolesForPressure (/obj/machinery/portable_atmospherics/proc/MolesForPressure)
  src: the Canister: \[O2] (/obj/machinery/portable_atmospherics/canister/oxygen)
  src.loc: the floor (83,100,1) (/turf/simulated/floor/tiled)
  call stack:
  the Canister: \[O2] (/obj/machinery/portable_atmospherics/canister/oxygen): MolesForPressure(14996.1)
  the Canister: \[O2] (/obj/machinery/portable_atmospherics/canister/oxygen): New(the floor (83,100,1) (/turf/simulated/floor/tiled))
[00:00:00] Runtime in canister.dm,405: Cannot execute null.adjust gas().
  proc name: New (/obj/machinery/portable_atmospherics/canister/oxygen/New)
  src: the Canister: \[O2] (/obj/machinery/portable_atmospherics/canister/oxygen)
  src.loc: the floor (83,100,1) (/turf/simulated/floor/tiled)
  call stack:
  the Canister: \[O2] (/obj/machinery/portable_atmospherics/canister/oxygen): New(the floor (83,100,1) (/turf/simulated/floor/tiled))`

PsyCommando:
Okay I think I fixed the damn runtime. And that stupid heat exchange proc running when there's no loc set on a canister..

code/game/machinery/atmoalter/portable_atmospherics.dm
```diff
@@ -144,6 +149,16 @@
 	var/last_power_draw = 0
 	var/obj/item/weapon/cell/cell
 
+/obj/machinery/portable_atmospherics/powered/Initialize()
+	. = ..()
+	if(!map_storage_loaded)
+		cell = make_cell()
```
Andrew-Fall: Making initialize call a proc like this seems like addition of spaghetti code. Instead it should use a var/const/spawn_cell = /obj/item/weapon/cell/apc and call new spawn_cell().
This would mean they just have to set the variable.

code/game/machinery/atmoalter/portable_atmospherics.dm
```diff
@@ -27,13 +23,22 @@
 	. = ..()
 
 /obj/machinery/portable_atmospherics/Initialize()
+	if(!air_contents)
+		init_air_content()
```
Andrew-Fall: Similar to below (minus the const because byond doesn't support that), this should iterate through a spawn_air list that has the gas and it's pressure list("phoron" = 1000) along with setting the temperature and updating the icon.

Andrew-Fall (dismissed):

code/game/machinery/atmoalter/portable_atmospherics.dm
Andrew-Fall: Apparently /const can't be overridden by children like regular code. In such a case just make spawn_cell a regular variable but do the rest.

PsyCommando: I tend to disagree. The advantage of having a proc creating the cell, is that you can do any kind of preparation you want inside the proc, unlike with a variable. Also, the variable is just taking up space in memory when 99% of its existence is for nothing, while the proc isn't since its only executed when needed.

code/game/machinery/atmoalter/portable_atmospherics.dm
PsyCommando: Its not that simple. Some just set the temperature and volume. Also, a list like that would have a lot of redundant information and take a lot of space in memory for literally nothing. Using a proc is much cleaner and less wasteful. Plus if in the future we need to add any extra steps to initializing gases this approach would make things significantly easier to edit and less tedious than editing 30 gas lists, AND adding extra code.

michaelchessall (approved):

PR #994, opened 2018-10-31
Un-commented the commented gun code

PsyCommando:
I know brawler spoke about needing to do this, so here.


ourpar:
if we want them obtainable to everyone we should add them back to the protolathe too, but if they're just going to be event items or something then ignore me

PsyCommando:
I'll try to make them buildable in a bit

Kurgis:
Someone made many of these guns build-able, that's the reason they were commented out. To delete them from the map without causing errors.

PsyCommando:
> 
> 
> Someone made many of these guns build-able, that's the reason they were commented out. To delete them from the map without causing errors.
Sounds good, I was just told by the others they didn't know why.

michaelchessall (approved):

PR #995, opened 2018-10-31
Aghost can now change job assignment

Chryseus:
Allows admin ghost to change job assignment and set custom title which is helpful during server setup.


code/modules/world_save/map_storage.dm
```diff
@@ -432,7 +432,6 @@ var/global/list/debug_data = list()
 
 /proc/Load_World()
 	var/starttime = REALTIMEOFDAY
-	if(!fexists("map_saves/game.sav")) return
```
Andrew-Fall: This probably shouldn't be changed...

code/modules/modular_computers/file_system/programs/command/card.dm
```diff
@@ -428,20 +428,29 @@
 						return 0
 				else
 					var/datum/computer_file/crew_record/record = connected_faction.get_record(user_id_card.registered_name)
-					var/datum/assignment/user_assignment = connected_faction.get_assignment(record.assignment_uid)
 					var/datum/assignment/assignment = locate(href_list["assign_target"])
 					if(!assignment) return 0
-					if(connected_faction.in_command(user_id_card.registered_name) || (user_assignment && user_assignment.parent.name == assignment.parent.name) || isleader)
+					if (!isghost(user))
+						var/datum/assignment/user_assignment = connected_faction.get_assignment(record.assignment_uid)
```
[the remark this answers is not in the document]
Andrew-Fall: Refering to the connected_faction.in_command(...

code/modules/world_save/map_storage.dm
Chryseus: It doesn't appear to ever be created, I've had to manually make it running the server to get loading to work.

PR #997, opened 2018-11-01
Fixes rezadone recipe caused by a revert

tacct6:
Rezadone now uses ryetalyn instead of carpotoxin once again.


Andrew-Fall (approved):
Code checks out, Travis passes.

PR #998, opened 2018-11-01
Fix pizza boxes from respawning pizza

Chryseus:


Andrew-Fall:
Although try to add a changelog next time. I forgot about those.

PR #1003, opened 2018-11-03
Update code/modules/power/cell.dm

UndeadNappist:
Makes potato cells and slime cores self-charging once again. Things have been tested and a cat doing zero division has been removed.

The charge is an extremely low trickle-charge, and the max capacity of these cells is insignificant compared to many others.

### DIFF
--- a/code/modules/power/cell.dm
+++ b/code/modules/power/cell.dm
@@ -263,7 +263,7 @@
 	icon = 'icons/obj/power.dmi' //'icons/obj/harvest.dmi'
 	icon_state = "potato_cell" //"potato_battery"
 	maxcharge = 20
-	const/self_recharge = 0.1
+	const/self_recharge = 0.2
 
 /obj/item/weapon/cell/slime
 	name = "charged slime core"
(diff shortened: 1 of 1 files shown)


code/modules/power/cell.dm
```diff
@@ -15,6 +15,10 @@
 	var/charge			                // Current charge
 	var/maxcharge = 1000 // Capacity in Wh
 	var/overlay_state
+	var/self_recharge = 0
+	var/charge_tick = 0
+	var/charge_tick_rate= 4
+	var/selfchargeamt = 0
```
Andrew-Fall: As I am fairly certain charge can have decimals, there should be only one constant variable for var/const/charge_rate = 0. If they want it to charge slowly, they can set it to 0.25 or something. Then Initialize will just check if charge_rate == 0 and Process would += charge_rate. This would also allow slow discharge for batteries assuming the coder wants that.

UndeadNappist: Changes made, although it's a cell with procs so it remains src.give

Made constants in the cells themselves.

UndeadNappist:
This will be a buff for a number of devices that have use potato cells but also have recharging built hardcoded into them, for inexplicable reasons.

This is limited to the defib and a number of other admin-spawn only items.

PR #1004, opened 2018-11-03
re-adds some lost clothing

tacct6:
Re-adds clothing sprites from https://github.com/Persistent-SS13/Persistent-Bay/commit/9ecdc9bec213b384f3a7622944900410cc4a6a99 that became lost sometime between then and now. 

It seems that PR https://github.com/Persistent-SS13/Persistent-Bay/pull/863 restored the code for these items but didn't restore the icons, so pretty much this PR restores the icons. Also the empty space to show arms when facing left and right were missing too, so I spent the time fixing that aswell. These _are_ clothing items that were already added, but if requested I can post a picture here showing them off.


Andrew-Fall (approved):
Code checks out and Travis passes.

PR #1006, opened 2018-11-03
Auto Save Pause

SirBloodBuske:


Andrew-Fall:
I unfortunately already have a SSautosave mostly complete that will handle most of this.

Andrew-Fall:
Also fails Travis due to use of world <<
You can use to_world() if needed.

SirBloodBuske:
> I unfortunately already have a SSautosave mostly complete that will handle most of this.

Okay, Sorry then, Would of been PR'd sooner but I been outa state for a bit, Ill just wait to see how your's acts then.

PR #1009, opened 2018-11-03
Admin ghost can now change job assignment

Chryseus:
Cleaned up from my previous PR.


Andrew-Fall:
Typically you'll want to provide a change-log as well, but they arn't automated anyways so I'll let it slide.

PR #1016, opened 2018-11-05
Add removable cells to energy weapons

PsyCommando:
After getting many requests, I decided to do it. And I added a lot more flexibility to help with balancing.

Energy weapons have now 3 ways of handling their cell:
* Fixed: Unremovable cell, like all energy guns right now.
* Hot-swappable: Basically like a ballistic weapon magazine.
* Removable: Also like a magazine, but you need to unscrew the cover to change the cell.

All energy weapons have a list of allowed cells types which can be changed as needed.

Also fixed cell/device/variable not reloading from a save properly. The constructor would receive a null value as max charge for the cell on reload. So I made it so it wouldn't affect a null value as max charge on new. There are no reasons to init the cell with a max charge of 0, so this makes things simpler.

Energy weapons are still compatible with rechargers. The change was made to have as little impact on how things works right now as possible.


michaelchessall (approved):

2cwldys:
I would very much like to see a lot more hot swappable energy weapon ammo capacities, so you have to use a recharger for the 'energy weapon magazine'

PR #1021, opened 2018-11-05
Fix shuttles

Chryseus:
Fixes shuttle creation and flight.


code/modules/shuttles/shuttle.dm
```diff
@@ -176,42 +177,62 @@
...
+	
 	for(var/area/A in shuttle_area)
-		// if there was a zlevel above our origin, erase our ceiling now we're leaving
+		// Remove ceiling we are leaving behind if it exists
 		if(HasAbove(current_location.z))
 			for(var/turf/TO in A.contents)
 				var/turf/TA = GetAbove(TO)
-				if(istype(TA, ceiling_type))
-					TA.ChangeTurf(get_base_turf_by_area(TA), 1, 1)
+				if(istype(TA, ceiling_type) || istype(TA, /turf/simulated/floor/airless) || istype(TA, /turf/simulated/open))
```
michaelchessall: This section is problematic as heck. I'm worried about shuttles landing underneath constructed areas and tearing them apart when they leave. This also deletes the contents of the shuttles roof, which is not a good idea, in any circumstance.

michaelchessall (changes requested):

code/modules/shuttles/shuttle.dm
Chryseus: You're right, I'll update it so it works properly in all situations.

Chryseus: I'm quite happy with how the shuttles work now, docks have a new variable for interior or exterior which should be saved, items left on the roof are now also kept, although it might make more sense for them to be knocked away when the shuttle moves.

Chryseus:
**Current major issues remaining**

- External docks need to check for non-space areas above to avoid causing damage.
- Have docking beacons reset properly when moved

**Features being worked on for a later PR**
- Allow arbitrary shuttle sizes (within reason)
- Rotation of shuttles
- Fuel consumption, internal / external fuel port, etc

Chryseus:
PR is ready for review again.
Extensive testing has been done and I can't get it to break so that's a good sign at least.

michaelchessall (approved):

PR #1024, opened 2018-11-06
Uncomments some viruses that were commented out.

Pandolphina:
:cl: Pandolphina
tweak: Uncomments a few Stage 4/Stage 3 viruses
/:cl:

Okey, reasoning time. First off, Stage 4's can't be obtained without admemes anyway so no reason to comment them out.

Second, the big evil bad random virus was removed.

Also All-Vax prevents "entire station dies from a cough that mutated the poison trait".

Look some people just want to be executed under the Space Geneva convention for making bioweapons. May be used as an NT sabotage as well.

### DIFF
--- a/code/modules/virus2/effect.dm
+++ b/code/modules/virus2/effect.dm
@@ -78,7 +78,6 @@
 				if(O)
 					O.droplimb(0,DROPLIMB_BLUNT)
 
-/**
 /datum/disease2/effect/radian
 	name = "Radian's Syndrome"
 	stage = 4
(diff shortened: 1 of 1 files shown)


FaurnsGoldberg:
uncomments killer viruses. not a great idea, despite all-vax. people will drop dead while AFK without much reason

Pandolphina:
The quick death viruses are admin-only, and the random Virus event is gone, I believe.

Plus, this adds actually use to virology.

PR #1025, opened 2018-11-06
Adds the Ammunition Fabricator.

Pandolphina:
:cl: Pandolphina
add: Adds the Ammunition Fabricator. Obtained by screwdrivering an autolathe, and is the only way to print ammo.
add: Bullets now cost chemicals. Normal ones cost aluminum, flash rounds cost aluminum and sulfur, and rubber rounds cost material plastic.
/:cl:

Please kill me. None of this worked as a child of the autolathe so I just went the smartfridge way.

### DIFF
--- a/code/game/machinery/ammofab.dm
+++ b/code/game/machinery/ammofab.dm
@@ -1,0 +1,8 @@
+/obj/machinery/autolathe/ammo_fab
+	name = "ammunition fabricator"
+	desc = "Fabricates many types of ammunition, magazines and boxes."
+	icon_state = "autolathe"
+
+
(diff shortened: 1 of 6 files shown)


PsyCommando:
Alright, so I was working on a ammo fabricator too. But if you wanna take over it I guess it might be fine, but I spotted some issues here.

Filling that huge ass recipe list in the ammo fab manually is bad practice. You should look at `autolathe_datums.dm`. They just make a list using the subtypes of `/datum/autolathe/recipe` in inside the `/proc/populate_lathe_recipes()` proc. So you can make your own datum class for the ammo fabricator and do the same. That's a lot more future proof and easier to maintain. So you won't have to ever add or remove anything from that list it'll be down automatically each times you add new recipes subtypes.

Also, personally, I would have made a separate circuit board for it, so it can be locked behind different requirements than the autolathe later on if needed. Considering the autolathe is pretty much something required for everyone to build, and ammo forges not so much.

code/game/machinery/ammofab.dm
```diff
@@ -0,0 +1,19 @@
+/obj/machinery/autolathe/ammo_fab
+	name = "ammunition fabricator"
+	desc = "Fabricates many types of ammunition, magazines and boxes."
+	icon_state = "autolathe"
+
+/obj/machinery/autolathe/New()
```
Andrew-Fall: Should be /obj/machinery/autolathe/ammo_fab/New()

code/modules/projectiles/ammunition/boxes.dm
```diff
@@ -7,7 +7,7 @@
 	icon_state = "38"
 	caliber = "357"
 	ammo_type = /obj/item/ammo_casing/a357
-	matter = list(DEFAULT_WALL_MATERIAL = 1260)
+	matter = list(DEFAULT_WALL_MATERIAL = 1260, /datum/reagent/aluminum = 30)//Speadloaders are less efficient, and should be balanced
```
Andrew-Fall: Use a tab to separate the comment from the code better, and Speedloaders*

code/modules/projectiles/ammunition/boxes.dm
```diff
@@ -17,7 +17,7 @@
 	icon_state = "38"
 	caliber = ".50"
 	ammo_type = /obj/item/ammo_casing/a50
-	matter = list(DEFAULT_WALL_MATERIAL = 1260)
+	matter = list(DEFAULT_WALL_MATERIAL = 1260, /datum/reagent/aluminum = 40)//The fucking instant kill device costs more
```
Andrew-Fall: Tab to seperate

Andrew-Fall:
Also please compile and test your code before creating a PR.

Pandolphina:
The code compiles, but in my sleep deprived state I did not test it. Will test and fix the things you mentioned in around 10 hours.

Pandolphina:
Okey, so, it's something with a map path error, atleast according to Travis. How do I fix this?

PR #1026, opened 2018-11-07
Made a check for map load before creating a new proxy assembly/air co…

PsyCommando:
…ntent

Should fix issues with tanks with a million assemblies in them. Just gotta make sure tanks save their air content in their saved var first.. Or crash might happen.


Andrew-Fall (approved):
Code checks out, awaiting Travis

PR #1042, opened 2018-11-12
Add bluespace canisters

Chryseus:
These are a convenient way to setup your own atmospherics systems, each one holds 200x the volume of a regular canister and can be ordered from supply for a much higher cost (may need tweaking), the sprites are functional but could use some improving, I also increased the health a bit as a rupture is pretty nasty.


Pandolphina:
Great idea, but what about when someone opens one in a crowded hallway at max pressure? They're kinda weapons of mass destruction.

Chryseus:
> 
> 
> Great idea, but what about when someone opens one in a crowded hallway at max pressure? They're kinda weapons of mass destruction.

I've added engineering access requirement to the valve and pressure setting, I also increased the health to 600 so it now takes quite a determined effort to break, all ordered tanks now have access requirement.

NataKilar:
Most of this code seems to be copy pasted from normal canisters - Is there any reason bluespace canisters couldn't be made a subtype, with different variables for health and volume?

Chryseus:
> 
> 
> Most of this code seems to be copy pasted from normal canisters - Is there any reason bluespace canisters couldn't be made a subtype, with different variables for health and volume?

That might be doable, I'll see what I can come up with.

Chryseus:
May not be needed so I'm closing this for now.

PR #1062, opened 2018-11-21
Add resource generation to asteroid turf

Chryseus:
This adds random resources to non-generated asteroids, amounts may need to be tweaked for optimal balance.

### DIFF
--- a/code/modules/mining/mine_turfs.dm
+++ b/code/modules/mining/mine_turfs.dm
@@ -462,14 +462,12 @@ var/list/mining_floors = list()
 	mining_floors["[src.z]"] += src
 	if(prob(70))
 		overlay_detail = "asteroid[rand(0,9)]"
-	message_admins("MSL: [map_storage_loaded], len [resources.len]")
 	if(map_storage_loaded != 1)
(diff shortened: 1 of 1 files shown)


michaelchessall:
Im always looking for help with asteroid code, but this actually is not the way we want asteroid turfs to populate. Currently we use 'oremaps' that apply ores across a zlevel creating pockets of high yield materials . In the next map we want to have ore generation that creates asteroids of entirely different compositions for players to find and contend with. Having asteroids generate a small smattering of low value materials on creation runs counter to this, as we may want some asteroids to even be bare/only sand.

Chryseus:
> 
> 
> Im always looking for help with asteroid code, but this actually is not the way we want asteroid turfs to populate. Currently we use 'oremaps' that apply ores across a zlevel creating pockets of high yield materials . In the next map we want to have ore generation that creates asteroids of entirely different compositions for players to find and contend with. Having asteroids generate a small smattering of low value materials on creation runs counter to this, as we may want some asteroids to even be bare/only sand.

I agree that sounds like a much better idea, this was more of a quick and dirty fix to test the mining system.

PR #1063, opened 2018-11-21
Adds a power requirement to ecigs,shortwave radio, and flashlights + changes to other handheld lighting

tacct6:
Ecig and shortwave radio code was (mostly) taken from baystation and ported over here
Power requirements/fuel amounts for each item are open for discussion if anyone disagrees with them, I was mainly focused on getting the items to simply require power!
* 1.Shortwave radios, flashlights, and ecigs now require device cells to operate
* 2.Flares, candles, and glowsticks have had their fuel amounts changed drastically
* 2.1: flares last between 15-30 minutes, candles last ~6 hours, and glowsticks last about 24 hours before becoming useless.
* 3.Some various light adjustments done to flashlights, mostly just making them brighter to give them a reason to use over other light-giving devices that don't (yet?) have a power requirement

Initially I was going to add power requirement to most of the handheld electronic devices but then I got distracted by other things and pretty much forgot about this, I figure PRing the code that I managed to finish so far would be better than just letting it disappear, so here it is!


tacct6:
i'll fix travis in a few hours

michaelchessall (approved):

PR #1065, opened 2018-11-24
Stackable ore

PsyCommando:
Alright, after much much testing, and many many issues, I think its good to go.
I got a bit carried away though on the way, and I did a few extra changes:

> * Fixed mining machines handling stackable ores
> * Fixed mining turfs dropping stacked ore.
> * Fixed typo in material name from "graphene" to "graphite". Graphene is a human produced material, while graphite is carbon naturally occuring.
> * Fixed titanium being in material limbo. Gave it an ore and added more details to its entry. Added it to the drill.
> * Replaced mineral console's "scan" with simply connection via multitool buffer. I couldn't figure why the scan didn't work and this is waaay easier and more consistent.
> * Made stackable ore overlay ore icons on top of itself as the amount of ores increases. Ideally a few new sprites could replace this.
> * Added to material datums the var "ore_matter" since ores usually contain a mix of several materials.
> * Renamed "tetrahedrite" to "freibergite" since the later contains a lot more chemical elements and materials.
> * Changed steel alloy recipe from "hematite + graphene"(?) to "iron + graphite"(graohite is carbon so it makes slightly more sense)
> * Added material data for aluminum, since we literally have bauxite already anyways. So alumium sheets can be smelted from bauxite.

Oh, and each ore stacks can contain up to 250 ores, which means, previously those 2,000 ores object we'd have on screen could as little as 8 stack objects if they were all the same type of ores. Should greatly improve performances with handling ore, and during mining.


PsyCommando:
Well, making this pr taught me I hate the oxygen generator! XD
Is is even like buildable?

michaelchessall (approved):

PR #1068, opened 2018-11-27
Added bombsuit crate to sec and science supply categories

PsyCommando:
* Added a pair of bomb resistant boots based off the swat boots to go with the suit.

Bombsuits were missing in the supply app, and they can be useful for security and science. So I added them.


Andrew-Fall (approved):
Code checks out, awaiting Travis

PR #1071, opened 2018-11-28
Update hydroponics tray right away

Chryseus:
This fixes the issue of slow updating of the hydroponics tray water and nutrient content, it also increases the update amount to 100 units per call rather than some small amount, I've tested it and there doesn't appear to be any side effects, not that I'd really expect any.


code/modules/hydroponics/trays/tray.dm
```diff
@@ -502,6 +498,12 @@
 		qdel(O)
 		check_health()
 
+	else if ( istype(O, /obj/item/weapon/reagent_containers) )
+		if( O.reagents.total_volume > 0 )
+			spawn(0)
+				sleep(5)
+				process_reagents() // Force reagents to be processed
```
Andrew-Fall, line 505: Should be able to call process_reagents right away. And spawning then sleeping is frowned upon. For most cases you should use AddTimer for anything relating to spawn/sleep.

Andrew-Fall (changes requested):

code/modules/hydroponics/trays/tray.dm
Chryseus: I tried that and it didn't work, adding a short delay fixed it so I could go with just a sleep, keep it as it is or try change it to use AddTimer, what would you recommend ?

Andrew-Fall: Not sure how this actually works, as hydroponics doesn't have any code for reagent_containers except for syringes and doesn't call ..()

Chryseus: The create_reagents() call in the tray initialization makes it a reagent container (adds the reagents var), when a reagent container is used it transfers its contents to the reagents var, that is then processed in process_reagents() when the controller updates the trays, my guess is the reagent transfer occurs after attackby() is called so process_reagents() gets called before it has a chance to complete the reagent transfer hence why the delay works.

Andrew-Fall: That would make sense. Yea just use AddTimer

Chryseus: A little more digging and it appears the reagent transfer is done in afterattack which is called after attackby, a call to handle_reactions() or update_total() might be an alternative to sleeping, I will have to check.

PR #1074, opened 2018-11-29
Cargo train power cell can now be removed

Chryseus:
Power cell can now be removed and replaced.


tacct6:
the code for removing/adding cells to cargo trains already exists here: https://github.com/Persistent-SS13/Persistent-Bay/blob/dev/code/modules/vehicles/vehicle.dm#L77

last I checked that code worked just fine when removing/adding cells, but that was a while ago, if something did break with them it _proobably_ should be fixed there

Chryseus:
> 
> 
> the code for removing/adding cells to cargo trains already exists here: https://github.com/Persistent-SS13/Persistent-Bay/blob/dev/code/modules/vehicles/vehicle.dm#L77
> 
> last I checked that code worked just fine when removing/adding cells, but that was a while ago, if something did break with them it _proobably_ should be fixed there

Ah I seem to have completely missed that, but yeah it isn't working so I'll get it fixed there.

Chryseus:
Well this is rather odd, it works perfectly fine now.

PR #1076, opened 2018-12-01
Auto-tailor + tons of clothing designs

tacct6:
**Features:**

1. 5 new autotailor machines, each with their own specific set of designs, you screwdriver the standard circuit printed from RD to access subtypes
2. Added empty variants to many storage containers, so the tailoring machine wouldn't print out toolboxes full of tools
3. Modified (in most cases, removed) armor values for many jumpsuit clothing items, usually clothing that were previously used for antags
4. cheap insulated gloves now have a static 0.2 shock absorption instead of RNG, so people don't keep printing cheap gloves from the tailor until they get a lucky pair (I'm hoping 0.2 is still too weak to be used as proper gloves, but strong enough to stop your brains from being fried out from a single mistake)
5. Added stethoscope (clothing accessory) to the protolathe, also uncommented the defibs since i noticed them existing
5. fixed & changed some mob sprites for various clothing (everything printed from the auto-tailor SHOULD have sprites)
6. tons of other small changes relating to clothes that would make this feature list too long if i noted them

Hopefully the big amount of additions isn't too scary, 90% of it is just simply adding designs as there are about 900 of them, some item names probably aren't too fitting compared to the tailoring designs (example: "blue shirt" might print out a captain jumpsuit), i'll most likely get around to renaming those before the new map hits


tacct6:
with adding so many designs i knew i was bound to add duplicates, fixing noow

michaelchessall (approved):

PR #1085, opened 2018-12-09
Adds functional support for vatgrown

tacct6:
Takes https://github.com/Persistent-SS13/Persistent-Bay/pull/818 and updated to be brought with current code, includes:

- Vatchips -  a special type of lace given to only vatgrown, these cannot be be brought back to life under any circumstances!

- They are nearly identical to humans, the only notable differences being paler skin and bleached hair 

- Names based off of baystation blanks, a Greek letter followed by some numbers/letters, (ex. Nu - E92), these names are completely random upon creation and cannot be selected by the player.

For now they are still unplayable until lore is finalized & made official, this also means any info in this PR shouldn't be considered official and could be changed in the future


tacct6:
since vatgrown aren't going to be making an appearance ingame with this PR I don't see much of a reason to make a changelog, any changes that do matter are already noted in the initial post

michaelchessall (approved):

PR #1087, opened 2018-12-09
Made logs accessible again

PsyCommando:
* Made logs accessible in-game again. Vebs used to give a file path error.
* Fixed logs being written to several arbitrary folders. They're now all in the "data/logs/YYYY/MM/DD/name.log" format/path
* Defined macros to get the path to the various logs.

Admin verbs weren't working since the logs were not written at the expected location. Standardized the log location using macros, and made sure everything writes at the same place now.


PsyCommando:
Will have to change travis stuff for this one..

PR #1090, opened 2018-12-29
GIT-HELL- Damage system refactor, Machinery refactor, Chemistry temperature port, etc

PsyCommando:
A big collection of fixes and tweaks to various thing to make them buildable/deconstructable, along with bug fixes and etc..
It took me about 2 weeks to do that, and also filter it from the mess that has become my local copy >_<

EDIT: Alright, so I was a bit foolish to even try to split this up after the damage was already done.
So I just merged all my PRs into this one.

Most of the changed files are mainly just single line changes for changing literals for constants, or replacing function calls. 

To sum up the changes here:

**Damage system refactor:**
 * Objects handle damages much more closely to how mobs do. And all child class of "/obj" inherit those base damage procs.
 * All objects can have armor, and damage thresholds.
 * All damage types are represented by clear constants, and you don't need to "translate" them from organ damage types or whatever other damage types. They work for nearly everything (except wounds.. and the current broken mech implementation.. for now)
 * Objects can be destroyed differently depending on damage type, and they can leave behind debris.
 * Ported some of the limb flags system bay upgraded to from using a ton of boolean vars. Which makes things a bit more closer to bay, and less convoluted.
 * Made glass windows a bit more resistant. So even the cheapest glass is still decent and won't explode after one or two punches.
 * Added the damageable object flag, so some objects can omit it to ignore damaging interactions.
 * All objects have a common damage type var associated with impact damage for objects, and attack damage for weapons. (Bullets still need some work to iron a few last things out)
 * Refunding materials is now integrated in the obj base class. It simply creates sheets from the content of the "matter" var. So its mostly automatic now.

**Machinery refactor:**
 * Added more procs to encapsulate and make several aspect of the machines polymorphic, Like turning the machine on and off, dismantling it, etc.. Mainly added a lot of convenience procs, constants and etc.
 * Ported power use constants from Bay. Pretty simple, but make things a whole lot clearer.
 * Added a radio transmitter extension to the machinery class. This allows to have all the radio stuff handled almost seamlessly, and consistently across all machines. Checking if a machine has a radio receiver is very straightforward too. Still need to be fully integrated, but seems to work alright so far.
*  Added broken state, so machines can enter broken state from reaching the specified health threshold, and reimplement the "broken" proc to add extra effects.

**Made many objects buildable**
 * Signs, several wall mounted machines, etc are all buildable.
 * Added tons of new recipes sorted by categories in the hand-crafting screen for materials. (Keep in mind I did that part long before we had the various fabricators...)
 * Several machines that weren't constructable, or deconstructible are now.

**Added Bay chem temperature system**
 * Added SStemperature from bay to handle atom temp.
 * Added hotplate and coldplate machines, with their sprites.
 * Reworked the UIs for the hotplate to use nanoui
 * Updated recipes that need to reach a certain temp on bay to behave the same way.
 * Added a few missing recipes from bay
 * Added reagents and recipes to make several types of modern gunpowder
 * Re-organized recipes into separate files labelled by categories.

**Misc:**
 * Made macros to define saved variables directly in the code, to make things clearer and extendable in the future.
 * Ported Bay's new computer/console sprites, which match better with the new walls.
 * Ported europa's sprites for the newscaster, atm, and etc since they match the perspective of the walls we have now, and look a lot less silly.
 * Restored rucksacks which were in the game files, but missing a bunch of things here and there.(Was probably lost in commits on bay somewhere)
 * Added several sound effects all public domain or CC. Credits for CC works and public domain ones are found in the ogg's tags.
 * Made lizards behave more like mices, so they don't shove you around, and you can pick them up.
 * Made bags able to mass pickup small mobs from the floor.
 * Modified the recycler so it can extract useful contents from more types of items, including reagents. (Waiting on SOM to integrate properly) Also now grinds non-human mobs.
 * Reworked the crematorium so it works better, and is a bit less of a deathtrap.
 * Reworked the morgue so it works and saves properly.
 * Added a WIP trash incinerator, that works much like the crematorium. 
 * Ported Bay's enveloppes
 * Updated icons for the paper shredder, shredded paper, and photocopier to those of Bay.
 * Made opened secure closets deconstructable.
 * Added loads of empty variants. 

**_EDIT2:_** Apparently my welder tool re balance/refactor got also included. So now welder tools have all efficiency values and welding things is handled by the tool, so welding times and fuel use can be controlled properly depending on the tool's efficiency, or even eventually user skills. I meant to turn it to tools datum but ehh at this point. I can always do that later.


ike709:
wew lad

PsyCommando:
Rip byond

PsyCommando:
Alright so seems like fixing that huge mess of macro name conflicts fixed most things. Just gotta fix multi-damage on objects, and get logs properly sorted out.

PsyCommando:
Alright so, I'm still trying to figure out a smart way to work out multi-damage weapons, like energy swords. Because I can do it 2 ways, one just go for cutting damage when its on, but then regular old cut resistant armor can protect against it, which doesn't make much sense. Or 2, I have the thing inflict secondary energy damage in addition to cut damage (cut damage is necessary for dismemberment to trigger)
One simple way to implement multi-damage types would be using a damage list. But I'm worried that creating lists for all weapons that would mostly contain a single damage type would be a big waste of memory and performance when most things have a single damage type that wouldn't require a list to store at all. 
I could also add a secondary damage var for type and force I guess.

PsyCommando:
Most of it seems to work now. Just gotta fix blast doors and buttons using radios.

PR #1091, opened 2018-12-29
Chemistry temperature

PsyCommando:
Port of bay's reagent temperature system. Also add the hotplate, and a few extra reagents and recipes from bay.
Also added a few recipes for modern gunpowder and etc..
It probably relies on things in the last PR.. Because I did a dumb.


PsyCommando:

PR #1092, opened 2018-12-29
Dam sys

PsyCommando:
This is the big one.. The damage system refactor. It took centuries to replace all those goddamn string literals for damage names and such ;_;
From my testing everything seems to work as usual. But my guess is we'll probably find issues with it. 
Also included the machinery refactor that was supposed to be in another PR by accident, so go me..


PsyCommando:

PR #1093, opened 2018-12-29
Tweak lizards

PsyCommando:
Makes lizard mobs stop shoving you around all the time. They now behave like mices. They can be carried like mices too (put together an in-hand sprite for it too), fixed their gibbing anim.


PsyCommando:

PsyCommando:

PR #1095, opened 2019-01-08
Removes NTSL

NataKilar:
Straight and simple port of https://github.com/Baystation12/Baystation12/pull/24017

Similar reasoning  - NTSL is unmaintained and seldom used, yet has large exploitative potential. From what I've gleaned, arbitrary execution of JS code is not really a concern on the client side, as we don't use Goonchat, but potential for abuse on the server side still exists.


Andrew-Fall:
692 << uses (expecting exactly 751)
Update the << uses in test/check-paths.sh

code/modules/reagents/reagent_containers/food/sandwich.dm:74:error: n_ceil: undefined proc

Also need to generate globals and update test/run-test.sh line 245

NataKilar:
Must have missed a file include in the commit. I'll get that done along with test path and global changes later today

PR #1098, opened 2019-01-15
Nexus Map

PsyCommando:
Mostly done with it. Still need to pipe most of the deck 3 core, and make sure its not leaaking. And well, a lot of devices need to be configured, and some area maybe furnished. For now it shouldn't be merged.

Also, for some reasons I get the feeling the vignette might have been somehow hardcoded in some subsystems, because I can't get multi-z to work for some reasons, and also pipes self-delete on atmos init. 
I decided to put this branch here so maybe anyone with an idea can take a look and speed things up?


Kurgis:
> Also, for some reasons I get the feeling the vignette might have been somehow hardcoded in some subsystems, because I can't get multi-z to work for some reasons, and also pipes self-delete on atmos init.

@PsyCommando Multi-z is handled by landmarks, open up a vignette.dmm, and you should see the landmark in question along with object path on the bottom left side of the map.

PR #1099, opened 2019-01-17
Ports picket signs from TG with minor changes.

Viole7:
- Ports picket signs from TG.
- Messages on signs cannot be rewritten as with vanilla picket signs.
- Signs are crafted using the general fabricator.
- Signs can be deconstructed with wirecutters, yielding less materials than the cost of a new sign.

I hope this will encourage role-play and economic turnover of signs.


Viole7:
Woops, my version of the general fabricator is behind..
Edit: Ok it should be fixed.

michaelchessall (approved):
looks good!

PR #1105, opened 2019-01-21
Addition of functional swords for officers and those of high breeding

PlasticCat:


PsyCommando:
what makes this sword any different than all the officer sword variants we got already?

PlasticCat:
> what makes this sword any different than all the officer sword variants we got already?

The ones in game are beyond useless in combat. The above is made on par with a machete in terms of combat effectiveness

PR #1118, opened 2019-03-02
Cleans up bike code and improves usability

NataKilar:
Cleans up bike code significantly and fixes some bugs.

* Rebalances bike crashing to be more forgiving speed wise, but makes it do more damage.
* Adds cruise control to bikes. While enabled, friction will not be applied, and you'll have to brake manually. However, you will be able to maintain a constant speed much easier.
* Adds UI action buttons to toggle engine and cruise control
* Adds a sound to the bike when moving (Taken from here https://freesound.org/people/soundjoao/sounds/325808/). Those with ambiance disabled in preferences will not hear it.
* Bikes tick much faster now, and are much smoother to ride. There was no detectable performance hit on my machine, but server side testing is encouraged.


code/modules/vehicles/bike.dm
```diff
@@ -381,14 +397,26 @@
 	overlays += image('icons/obj/bike.dmi', "bike_overlay", "layer" = VEHICLE_TOP_LAYER)
```
PsyCommando: Better use src.icon as the first parameter for this. It avoids having to change the icon paths everywhere.

NataKilar: Addressed.

PR #1123, opened 2019-04-01
Fix and cleanup of SSmazemap changes from #1121

ingles98:
Fixes an issue in which even non-player mobs would trigger zlevel activation (Maze_map)

Also cleans an useless if statement away.

(This is a follow up for PR #1121)


code/game/turfs/turf.dm
```diff
@@ -126,7 +126,7 @@ var/const/enterloopsanity = 100
 
 	if(ismob(A))
 		var/mob/M = A
-		if (isliving(M) && M.stat != DEAD)
+		if (isliving(M) && M.stat != DEAD && M.client)
```
ingles98, line 129: Forgot checking if the mob is a player controlled one as well.

We dont want dolls activating the zlevels.

code/controllers/subsystems/SSmaze_map.dm
```diff
@@ -84,8 +84,7 @@ SUBSYSTEM_DEF(mazemap)
 		var/mob/M = C.mob
 		if (!M.z || !isliving(M) || M.stat == DEAD)
 			continue
-		if(activity_checklist["[M.z]"])
```
ingles98, line 87: This is true no matter what.

code/modules/maze_map/maze_map.dm
```diff
@@ -85,6 +85,8 @@ GLOBAL_LIST_EMPTY(maze_map_data)
 			STOP_PROCESSING(SSmobs, m)
 
 /datum/zlevel_data/proc/on_active()
+	if (!isWild())
+		return
```
ingles98, line 89: Needs this check in order to not replenish_monsters() when a safe zlevel is set to active.
The replenish_monsters() does not check for wilderness nor whether the zlevel is active, we do it apriori.

PR #1124, opened 2019-04-02
General Code Cleanup

Andrew-Fall:


Andrew-Fall:
Ugh gosh darn special characters

PR #1127, opened 2019-04-02
General fixes

Andrew-Fall:


code/game/antagonist/alien/borer.dm
```diff
@@ -38,7 +38,7 @@ var/datum/antagonist/xenos/borer/borers
 		for(var/mob/living/carbon/human/H in SSmobs.mob_list)
 			if(H.stat != DEAD && !H.has_brain_worms())
 				var/obj/item/organ/external/head = H.get_organ(BP_HEAD)
-				if(head && !(head.robotic >= ORGAN_ROBOT))
+				if(head && !(head.status & ORGAN_ROBOTIC))
```
PsyCommando: There's actually a macro for this BP_IS_ROBOTIC() Its what bay does and its more flexible than always doing the flag check

code/datums/extensions/extensions.dm
```diff
@@ -27,7 +27,7 @@
 				var/datum/extension/ext = extension
 				ext.set_holder(src) //Ensure the holder is set properly
 			else
-				log_debug("/datum/after_load(): found a list extension.. Not setting holder.")
+				log_debug(" /datum/after_load(): found a list extension.. Not setting holder.")
```
PsyCommando, line 30: Why are you adding a single white space in all those files?

maps/nexus/nexus_empty.dmm
```diff
@@ -1,4 +1,9 @@
 "a" = (/turf/space,/area/space)
+"b" = (/turf/unsimulated/wall,/area/space)
```
PsyCommando, line 2: Isn't that meant to be an empty map?

PsyCommando (changes requested):
Just a few things I noted.

code/datums/extensions/extensions.dm
ingles98: Looks automated, idk what could be the reason tbh

code/modules/flufftext/TextFilters.dm
```diff
@@ -134,14 +134,14 @@ proc/RadioChat(mob/living/user, message, distortion_chance = 60, distortion_spee
 					if(english_only)
 						newletter += "*"
 					else
-						newletter = pick("�", "�", "%", "�", "�")
+						newletter = pick("ø", "Ð", "%", "æ", "µ")
```
ingles98, line 137: Not entirely sure about what is going on with these but seem harmless enough

code/modules/mob/living/carbon/human/species/station/nabber.dm
```diff
@@ -226,7 +226,7 @@
 				if(part)
 					image_key += "[part.species.get_race_key(part.owner)]"
 					image_key += "[part.dna.GetUIState(DNA_UI_GENDER)]"
-				if(part.robotic >= ORGAN_ROBOT)
+				if(partstatus & ORGAN_ROBOTIC)
```
ingles98: Oof i believe this is supposed to be part.status am i right?

code/datums/extensions/extensions.dm
Andrew-Fall: Because Travis identifies those as text datum paths. Adding the space prevents that. Usually it indicates a text2path

maps/nexus/nexus_empty.dmm
Andrew-Fall: Yes minus admin z which doesn't save and is used for unit tests

code/game/antagonist/alien/borer.dm
Andrew-Fall: Ugh, I'll update them then.

ingles98 (approved):

code/modules/mob/living/carbon/human/species/station/nabber.dm
Andrew-Fall: Yep, that file isn't included in the build hence me missing it

PR #1129, opened 2019-04-05
Fixed some obvious errors in the damage system

PsyCommando:
* Turns out some parameters were wrong for a lot of damage proc calls. So I fixed them. Hopefully this helps with the damage issues.


Andrew-Fall (approved):
Everything checks out and passes Travis

PR #1130, opened 2019-04-05
Faction spawners didn't work, but now they do!

PsyCommando:
* Used lazy macros to handle dealing with the faction list.
* Fixed the faction spawners deleting themselves before spawning the actual faction.


code/modules/factions/bluespace_satellite.dm
```diff
@@ -609,7 +609,7 @@ GLOBAL_LIST_EMPTY(all_docking_beacons)
 			contract.finalize()
 			signed_contracts -= contract
 		var/datum/world_faction/new_faction = new()
-		GLOB.all_world_factions |= new_faction
+		LAZYADD(GLOB.all_world_factions, new_faction)
```
Andrew-Fall: Shouldn't these be LAZYDISTINCTADD?

Andrew-Fall (changes requested):

code/modules/factions/bluespace_satellite.dm
PsyCommando: I guess it could be. Its not like the list is associative or that theres a chance for newly made factions to be duplicates though. since its a simple array.

code/modules/factions/faction.dm
```diff
@@ -1382,9 +1383,12 @@ var/PriorityQueue/all_feeds
 	var/network_password
 	var/network_invisible = FALSE
 
-/obj/faction_spawner/New()
-	if(!GLOB.all_world_factions)
-		GLOB.all_world_factions = list()
+/obj/faction_spawner/Initialize()
+	. = ..()
+	spawn_faction()
+	qdel_self()
```
Andrew-Fall: Should be return INITIALIZE_HINT_QDEL

Andrew-Fall (changes requested):

code/modules/factions/faction.dm
PsyCommando: True

Andrew-Fall (approved):
Seems fine and passes Travis

PR #1190, opened 2019-06-23
fix missing sprites

tacct6:
bay merge changed where the game looks for clothing, most of our persistence specific clothing didn't make it, this moves the missing sprites to the correct location


PsyCommando:
Don't do that. I got something for that already. Moving sprites over is gonna do more bad than good. We wanna keep the bay base ones mostly the same and add new files with the things we added. Because otherwise things are gonna get replaced accidentally over and over again.

PsyCommando (commented):
Most of those are fixed in my pr already.

tacct6:
i'll close this one then.

PR #1198, opened 2019-06-24
Bug fiesta

PsyCommando:
A large collection of bug and runtimes fixes
* Fix many wrongly assigned icon for clothing


code/controllers/employment_controller.dm
```diff
@@ -31,10 +31,9 @@ var/datum/controller/employment_controller/employment_controller
 				else
 					employer:unpaid["[employee.real_name]"]++
 
+	var/list/paydata = list()
 	if(payday)
```
Andrew-Fall: Pretty sure this is an issue with tabbing below, where people themselves are payed

code/controllers/master.dm
```diff
@@ -223,11 +223,11 @@ GLOBAL_REAL(Master, /datum/controller/master) = new
 	for(var/mob/new_player/player in GLOB.player_list)
 		if(player.panel)
 			player.panel.close()
-			player.new_player_panel()
-			if(player && player.ready && player.mind)
-				player.loadCharacter()
-			else
-				message_admins("skipping player [player], [player.ready], [player.mind]")
+		player.new_player_panel()
+		if(player && player.ready && player.mind)
```
Andrew-Fall, line 217: We call player.new_player_panel(). Why check for player?

code/controllers/subsystems/initialization/character_setup.dm
```diff
@@ -13,24 +13,27 @@ SUBSYSTEM_DEF(character_setup)
 	var/list/save_queue = list()
 
 /datum/controller/subsystem/character_setup/Initialize()
-	if(newplayers_requiring_init.len)
+	wait = 3 //Tick faster at first to handle logging in players
+	. = ..()
+	//In 10 seconds slow down to once a second
+	spawn(10 SECONDS)
+		wait = 1 SECOND
```
Andrew-Fall, line 20: Might be better to put this in fire, such that once all are initialized, wait goes up to 1 second. This way it rapidly fires as much as it needs at the start.

code/game/machinery/cryopod.dm
```diff
@@ -55,9 +55,6 @@
 
 /obj/machinery/cryopod/SetupParts()
 	announce = new /obj/item/device/radio/intercom(src)
-	LAZYADD(component_parts, new /obj/item/weapon/stock_parts/matter_bin(src))
-	LAZYADD(component_parts, new /obj/item/weapon/stock_parts/scanning_module(src))
-	LAZYADD(component_parts, new /obj/item/weapon/stock_parts/console_screen(src))
```
Andrew-Fall, line 60: Was this and the other removals of part addition handled somewhere else?

code/modules/construction/girderparts.dm
```diff
@@ -12,7 +12,6 @@
 /obj/item/girderpart/New(var/newloc, var/mat)
 	..()
 	material = mat
-	name = "[material.display_name] " + initial(name)
```
Andrew-Fall, line 15: Why was this removed?

code/modules/factions/faction.dm
```diff
@@ -1,4 +1,4 @@
-GLOBAL_LIST_EMPTY(all_world_factions)
+GLOBAL_RAW(/list/datum/world_faction/all_world_factions); GLOBAL_UNMANAGED(X, InitValue); //Don't init as empty list, because it happens too late during init
```
Andrew-Fall: Change this to a standard var/global/list/all_world_factions = list()

Andrew-Fall (changes requested):

code/controllers/employment_controller.dm
PsyCommando: Brawler told me to move it there so the game would compile

code/modules/factions/faction.dm
Andrew-Fall: Test

Andrew-Fall (dismissed):

Andrew-Fall (commented):
Github why you no work

code/controllers/master.dm
```diff
@@ -213,11 +213,11 @@ GLOBAL_REAL(Master, /datum/controller/master) = new
...
```
Andrew-Fall, line 217: No need for if(player ...) as player will not be null

code/modules/factions/faction.dm
Andrew-Fall: There is a better way to do this I believe. I'll look it up

Andrew-Fall (changes requested):

code/controllers/master.dm
PsyCommando: Yeah, but I only un-indented this so it wasn't all under the if and wouldn't run at all.

code/modules/factions/faction.dm
PsyCommando: Yeah, writing a SSfaction of some sort. But I don't really have the time or will to go that far XD And we got enough bugs already ^^;

Andrew-Fall:
## UNIT_TEST ##: [UNIQUENESS: Research Designs Shall Be Unique]: Build Paths - /obj/item/weapon/circuitboard/optable - The following entries have the same value: /datum/design/circuit/operatingtable and /datum/design/circuit/operating_table

code/modules/factions/faction.dm
Andrew-Fall: Global Unmanaged doesn't work like that. In the define it's set to the values used in the macro. Use GLOBAL_RAW_VAR(blah...) = list() instead

PsyCommando: `#define GLOBAL_UNMANAGED(X, InitValue) /datum/controller/global_vars/proc/InitGlobal##X()
`
Well, it seems like it defines an empty proc. Also, I don't actually need/want it initialized, because I want it lazy accessed.

PsyCommando:
I'll make another one for removing the duplicate I guess

PR #1203, opened 2019-06-25
reimplements graphene grinding, adds a production method

Mazianni:
🆑 
Reintroduces graphene. Graphene is made by cracking Co2 in an oxygen regenerator. Graphene can be ground into carbon and acetone.
By the request of @ Mama
🆑

### DIFF
--- a/code/modules/materials/definitions/materials_mineral.dm
+++ b/code/modules/materials/definitions/materials_mineral.dm
@@ -50,8 +50,8 @@
 	stack_type = /obj/item/stack/material/carbon
 	icon_colour = "#141414"
 	chem_products = list(
-		/datum/reagent/carbon = 20,
-		/datum/reagent/acetone = 20
+		/datum/reagent/carbon = 10,
+		/datum/reagent/acetone = 10
 		)
 	sale_price = 1
 	ore_matter = list(MATERIAL_GRAPHITE = 2000)
--- a/code/modules/materials/definitions/materials_mineral.dm
+++ b/code/modules/materials/definitions/materials_mineral.dm
@@ -43,6 +43,20 @@
 	ore_matter = list(MATERIAL_GRAPHITE = 2000)
 	energy_combustion = 32.7
 
+/material/graphene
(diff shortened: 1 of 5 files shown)


PsyCommando (changes requested):
You really should use graphite for that?

code/modules/materials/definitions/materials_mineral.dm
```diff
@@ -43,6 +43,20 @@
 	ore_matter = list(MATERIAL_GRAPHITE = 2000)
 	energy_combustion = 32.7
 
+/material/graphene
+	name = MATERIAL_GRAPHENE //Aka carbon
```
PsyCommando, line 47: Graphite is raw carbon, and its made out of graphene layers.

Mazianni:
I don't think graphite's a good option, as literally everybody can get it.

This provides a production path that's one-step, through engineering companies or even government atmos.

PsyCommando (commented):
> 
> 
> I don't think graphite's a good option, as literally everybody can get it.
> 
> This provides a production path that's one-step, through engineering companies or even government atmos.

Well, how are people supposed to get it then? Acetone was meant to be easily obtainable as well as anything else, but we forgot about it.

Mazianni:
Graphene is made from Co2 crackers. They are machines, constructable, even, which turn Co2 into hot oxygen and graphene.

PsyCommando:
> 
> 
> Graphene is made from Co2 crackers. They are machines, constructable, even, which turn Co2 into hot oxygen and graphene.

Yeah, but then you can basically make it for free from waste gases you can get anywhere. Instead of having to get it mined.

Mazianni:
I've already implemented another chemical path that allows for acetone production, which already uses graphite. Grind graphite, combine with O2/H2, get acetone. 

The gas generator PR is related so people can get O2/H2 in reagent forms.

I understand why you'd want it to be graphite, but I really think we should be encouraging interaction, and this is one way - atmos is government controlled, so the gov can potentially produce acetone and sell it, or use it in a hospital.

PR #1205, opened 2019-06-26
UI when clicking mirrors now appear again

tacct6:
the fix is directly taken from baycode, where they had recently fixed the exact same issue. I also removed starting items from mirrors since deconstructing/reconstructing would yield these items infinitely if taken out.

for clarification the UI with mirrors allows people to change their hair/facial hair style & color, that's about it.


Andrew-Fall (approved):
Code checks out. Travis passes

PR #1208, opened 2019-06-26
Shovels can now dig sand.

Mazianni:
tl;dr usable-tools was coming first, and it assumed you wanted to dig for ores.
Not so.


Andrew-Fall (approved):
Looks good. Awaiting Travis.

PR #1212, opened 2019-06-27
Cryo_crate Patch 1 + Extras

Viole7:
*Changes crates to only spawn the contents upon opening.
*Possible fix to icon issue.
*Unsealed crates located on inactive z-levels delete after some time.
*Adds, removes, and re-balances some seeds in rarity.dm.
*Makes bees and bee-keeping gear obtainable. 
*Minor improvements and alterations.


code/game/objects/structures/crates_lockers/closets/cryo_crate.dm
```diff
@@ -42,17 +42,12 @@
 /obj/structure/closet/crate/cryo/New()
 	if(sealed)
 		content_path = select_cont()
-
-	if(content_path)
-		for(var/i = 1, i <= amount, i++)
-			var/mob/cold_critt = new content_path(src)
-			if(cold_critt in SSmobs.mob_list)
-				STOP_PROCESSING(SSmobs, cold_critt)
+
```
Andrew-Fall: There's 4 spaces in a row in this area instead of a tab

Andrew-Fall (changes requested):

code/game/objects/structures/crates_lockers/closets/cryo_crate.dm
Viole7: No clue how that got there, thanks for spotting it.

PsyCommando:
@Viole7 Just wondering, but can we find bees for bee hives in those crates?

Viole7:
> 
> 
> @Viole7 Just wondering, but can we find bees for bee hives in those crates?

Not currently, but I don't see why they couldn't be added. I'll add them tomorrow along with some minor changes I want to make. (Done)

PsyCommando (approved):
Looks good to me! Appearence probably won't work, but we really need this change for tomorrow.

Viole7:
> 
> 
> Looks good to me! Appearence probably won't work, but we really need this change for tomorrow.

I tried creating an entry in the appearance datum, the closed state worked fine but for some reason the open state wouldn't display. I used VV verify the state was being assigned correctly, manually setting the icon var caused it to display properly. This despite both states residing in the same .dmi file, so it shouldn't have mattered. Not exactly sure what was going on. I saw that other objects were setting closet_appearance to null and manually assigning the icons, so I went with that. Is it the server saving that's assigning the icon to the default in the appearance datum? If so, what exactly is doing that? I've never actually setup my test server to save, so I've only been testing this stuff without saves.

PR #1213, opened 2019-06-27
Gas Generator PR Mk2

Mazianni:
🆑 
Gas generator is constructable
Gas generator is desconstructable
Circuit board exists.
it's in the engifab.
🆑 

Now with 100% less gitfuckery.


PsyCommando (approved):
Looks good to me!

PR #1214, opened 2019-06-27
Make saved z-levels read from a list instead of a constant number

PsyCommando:
Basically a convenience tool to add/remove saved z-level to the list. So if we want to save an extra z-level like the admin z-levels, we can do it at runtime and not have to recompile each times. Or if we create new z-levels at runtime we can also add it. The constant number used to be impossible to change at runtime.


PsyCommando:
Still gotta test this properly..

PsyCommando:
One of the bit issue I got with this thing is that its really hard to get the z-level list to be saved. So even if you add a new z-level to the list, it won't be loaded on next map load. And my attempts to save it to a global list didn't work out too well.

PsyCommando:
I think I'll go back to the drawing board

PR #1222, opened 2019-06-28
hydroponic trays no longer disintegrate

Mazianni:
Why was this var changed.

Whyyyy.


Andrew-Fall (approved):
Code checks out and passes Travis.

PR #1238, opened 2019-06-29
Honey extractor can be deconstructed and built. + added circuit

PsyCommando:


Andrew-Fall:
Just need to update << count cause brawler's a bully and doesn't 😭

PsyCommando:
oof

PR #1245, opened 2019-06-30
Added more air tanks types to fab + matched their recipes to drops

PsyCommando:
* added rescue bag to medical fab (basically similar to cryobags and etc)
* You can now build most air tanks and fuel tanks


code/modules/research/fabricators/medical_fabricator.dm
```diff
@@ -66,6 +66,10 @@
 	build_path = /obj/item/bodybag/cryobag
 	materials = list(MATERIAL_CLOTH = 2 SHEETS, MATERIAL_PHORON = 1 SHEET)
 
+/datum/design/item/medicalfab/meditools/adv/rescuebag
+	build_path = /obj/item/bodybag/rescue
+	materials = list(MATERIAL_CLOTH = 2 SHEETS, MATERIAL_PHORON = 1 SHEET)
```
NataKilar, line 71: Seems a bit overpriced since rescue bags have no stasis function. Maybe use a metal instead of phoron?

PsyCommando: I literally just copy-pasted what brawler put for stasis bags honestly. But I can make it less expensive.

PR #1249, opened 2019-07-01
fabricators remember research datums now

Mazianni:
see title

I dunno if this is all i need to do, pls comment if not.


michaelchessall:
not actually the correct fix. In the circuit of the fabricator theirs a var called "unlocked_techs". I've already made that save.

PR #1255, opened 2019-07-01
Minable ices

PsyCommando:
Added a way to get some of the unobtainable/hard to get base reagents in the reagent form. Ices are basically ores that occurs along regular ore, and can be grinded into things like hydrogen, nitrogen, carbon dioxide, oxygen, acetone, ammonia, etc..

Added recipes for making welder fuel, plastic, etc.. Along with a few more reagents with their own properties. 
Also made some fixes to stack comparison.


Andrew-Fall (approved):
Checks out and passes Travis.

PR #1256, opened 2019-07-01
[WAIT ON BALANCE]Adds alternative precursor reagent recipes

NataKilar:
A few alternative recipes to get those all important precursor reagents. Doesn't replace the other ways of obtaining them through gas conversion or https://github.com/Persistent-SS13/Persistent-Bay/pull/1255, but is a bit more intuitive and available, even if it's not particularly realistic


PsyCommando:
I don't think you can actually get chlorine in reagent form. Also, that's a whole lot of influx of base chems suddenly with the reagents ices and all.

NataKilar:
Seemed to think you could through salt grinding or something but I think you're correct. I'll leave this as DNM for now and if players are still struggling with the amount of base chemicals we can reconsider it?

PsyCommando:
> 
> 
> Seemed to think you could through salt grinding or something but I think you're correct. I'll leave this as DNM for now and if players are still struggling with the amount of base chemicals we can reconsider it?

alright

PR #1257, opened 2019-07-01
Makes medical Stacks (ATK, BTK, Ointment, Gauze) producible

NataKilar:
These are necessary for proper medical gameplay.


PsyCommando:
Holy crap those costs o_0

NataKilar:
Should be noted that these create stacks, not single amounts.

PR #1272, opened 2019-07-03
Made many more programs compatible with tablets, laptops, and PDAs an…

PsyCommando:
…d reduced their sizes a little

* The PDAs and tablets were nearly useless because you couldn't run a lot of business programs on it and other basic things. This should help a bit


PsyCommando:
Apparently I put my clonepod fixes in here too.. So there's that

PR #1278, opened 2019-07-04
Egg reduction and half chicken meat/Seed extractor rng

HbsRed:


code/modules/mining/mine_items.dm
```diff
@@ -110,7 +110,7 @@
 		overlays += overlay_image(icon, "pick_handle", flags=RESET_COLOR)
 	. = ..()
 
-/obj/item/weapon/pickaxe/hammer
+//obj/item/weapon/pickaxe/hammer
```
Andrew-Fall: Thought Brawler fixed this

Andrew-Fall: This isn't using the /* */

Andrew-Fall (changes requested):

Andrew-Fall (approved):
Code checks out, waiting for Travis to be updated.

PsyCommando:
Its the dev branch again though

PsyCommando:
Dude, you're pushing more unrelated changes into your pr right now.. You really need to make a branch for your feature changes ^^;

code/game/machinery/seed_extractor.dm
```diff
@@ -30,7 +30,10 @@ obj/machinery/seed_extractor/attackby(var/obj/item/O as obj, var/mob/user as mob
 
 		if(new_seed_type)
 			to_chat(user, "<span class='notice'>You extract some seeds from [O].</span>")
-			var/produce = rand(1,4)
+			if(prob(35))
+				var/produce =(1)
```
PsyCommando, line 34: The problem with nerfing things this way is that starting player are enormously disadvantaged since they don't have much seeds to waste, and will get seedless at one point from plants dying, and bugs, pest, weeds, etc..
I'd suggest reducing the default plants yield instead and maybe make some of them more finicky. And maybe changing the extracted seeds to rand(1,2). So that you're not inherently fucked by the RNG.

rhicora: This is absolutely correct.

PR #1286, opened 2019-07-04
Materials

### LINKED ISSUE #1232: Material Rod Dupe
Material Rods (aka steel rods for grills) can be used to produce infinite steel via inserting them into a fabricator as each one gives one sheet worth of steel, which can be converted into two steel bars. This issues arises from material stacks having a separate var from materials (datum) that dictate how much material an object should have. (Something like perobj)

### LINKED ISSUE #1234: Low Walls don't respect Materials.
Low walls don't respect the material used nor to they yield them when deconstructed.

### LINKED ISSUE #1235: Girders parts are always listed as steel girder parts.
Title

### LINKED ISSUE #1284: Reinforced Walls
Produce sheets instead of rods when deconstructing.
Doesn't remove reinf_material creating infinite materials.

Andrew-Fall:
Closes #1284 
Closes #1234 
Closes #1235 
Closes #1232

### DIFF
--- a/code/modules/materials/recipes_stacks.dm
+++ b/code/modules/materials/recipes_stacks.dm
@@ -9,7 +9,9 @@
 
 /datum/stack_recipe/rod/spawn_result(user, location, amount)
 	var/obj/item/stack/S = new result_type(location, amount, use_material)
-	S.add_to_stacks(user, 1)
+	spawn(0)
+		S.add_to_stacks(user, TRUE)
+	
 	return S
 
 // Tiles
--- a/code/modules/research/fabricator.dm
+++ b/code/modules/research/fabricator.dm
@@ -1,3 +1,6 @@
+#define EJECT_ALL -1
+#define EJECT_ONE 0
+
 /*
 Fabricators
 A reworked and modular system intended to differentiate the production of items from RnD through specialized machines, in addition to giving them a nicer
@@ -106,7 +109,7 @@ as their designs, in a single .dm file. voidsuit_fabricator.dm is an entirely co
 
 /obj/machinery/fabricator/dismantle()
 	for(var/f in materials)
-		eject_materials(f, -1)
+		eject_materials(f, EJECT_ALL)
 	..()
 
 /obj/machinery/fabricator/RefreshParts()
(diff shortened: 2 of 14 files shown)


code/modules/materials/recipes_stacks.dm
```diff
@@ -9,7 +9,9 @@
 
 /datum/stack_recipe/rod/spawn_result(user, location, amount)
 	var/obj/item/stack/S = new result_type(location, amount, use_material)
-	S.add_to_stacks(user, 1)
+	spawn(0)
```
PsyCommando, line 12: I'm not sure why you're using spawn here? Sounds like a bit of a patch.

code/modules/research/fabricator.dm
```diff
@@ -298,26 +301,39 @@ as their designs, in a single .dm file. voidsuit_fabricator.dm is an entirely co
...
-			if(metal_load_anim)
-				overlays += "fab-load-[material]"
-				spawn(10)
-					overlays -= "fab-load-[material]"
-			while(materials[material] + amnt <= res_max_amount && stack.amount >= 1)
-				materials[material] += amnt
-				stack.use(1)
-				count++
-			to_chat(user, "You insert [count] [count == 1 ? stack_singular : stack_plural] into the fabricator.")// 0 steel sheets, 1 steel sheet, 2 steel sheets, etc
+	var/stackMatter = stack.material.get_matter()
```
PsyCommando, line 309: Normally it should be fine to just get the matter amount straight from the stack. Since stacks recalculate their amount of matter whenever something is added/removed so the calculation isn't duplicated everywhere and done over and over again externally.

code/modules/research/fabricator.dm
```diff
@@ -298,26 +301,39 @@ as their designs, in a single .dm file. voidsuit_fabricator.dm is an entirely co
...
-	var/amnt = stack.perunit
 
 	if(!(material in materials))
 		to_chat(user, "<span class=warning>\The [src] does not accept [stack_plural]!</span>")
 		return
 
-	if(materials[material] + amnt <= res_max_amount)
-		if(stack && stack.amount >= 1)
-			var/count = 0
-			if(metal_load_anim)
```
PsyCommando, line 310: Seems like you forgot to recopy the insert animation code here.

PsyCommando (changes requested):
Overall it looks really good! You've pretty much removed a lot of weird crud that accumulated in the stacks code. And I like what you did with the stacks constructor and saving the material name directly with read/write.

Only thing I'd really change maybe is the recalculation of the matter content for the stack inserted in the fabricator. 
And well, you probably want to add back the removed insert animation code.

code/modules/materials/recipes_stacks.dm
Andrew-Fall: It is, for some reason it won't qdel correctly if the code continues. I messed around with it for a while before giving up and resorting to that

code/modules/research/fabricator.dm
Andrew-Fall: I wanted the per-sheet matter to check if each sheet can fit. Some might have more then one material (aka reinforced materials)

code/modules/research/fabricator.dm
Andrew-Fall: Readded

code/modules/research/fabricator.dm
PsyCommando: Good call

code/modules/materials/recipes_stacks.dm
PsyCommando: That's interesting, because we've been having lots of issues with the GC being an ass for a while now..

PsyCommando (approved):

PR #1290, opened 2019-07-05
Adds new phorostimulant and some more addictions

rhicora:
- Adds phorostimulant (sugar + phoron)
- Adjusts alcohol to be slightly more forgiving
- Adds some more messaging to alcohol
- Adds mushroom trip messages
- Adds dependence values for phorostim and mushrooms


Andrew-Fall:
image

PR #1298, opened 2019-07-06
fixes ore processor and other such things

Mazianni:
so it turns out our ore processor was missing a line for whatever reason and couldn't use the bay recipes.

Fuck if i know why but this should more or less fix OCP/plasteel being unobtainable through non-silly means.

### DIFF
--- a/code/modules/mining/machinery/mineral_processor.dm
+++ b/code/modules/mining/machinery/mineral_processor.dm
@@ -47,7 +47,7 @@
 	if(input_turf)
 		for(var/obj/item/stack/S in input_turf)
 			// If we are deleted or are neither dust or ore, continue
-			if(QDELETED(S) || (!istype(S, /obj/item/stack/material_dust) && !istype(S, /obj/item/stack/ore) && !istype(S, /obj/item/stack/material) || !LAZYLEN(S.matter))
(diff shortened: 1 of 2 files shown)


Andrew-Fall:
image

code/modules/mining/machinery/mineral_processor.dm
```diff
@@ -47,7 +47,7 @@
 	if(input_turf)
 		for(var/obj/item/stack/S in input_turf)
 			// If we are deleted or are neither dust or ore, continue
-			if(QDELETED(S) || (!istype(S, /obj/item/stack/material_dust) && !istype(S, /obj/item/stack/ore)) || !LAZYLEN(S.matter))
+			if(QDELETED(S) || (!istype(S, /obj/item/stack/material_dust) && !istype(S, /obj/item/stack/ore) && !istype(S, /obj/item/stack/material) || !LAZYLEN(S.matter))
```
ike709:
```suggestion
			if(QDELETED(S) || (!istype(S, /obj/item/stack/material_dust) && !istype(S, /obj/item/stack/ore) && !istype(S, /obj/item/stack/material)) || !LAZYLEN(S.matter))
```

ike709: @Andrew-Fall Is adding a code suggestion to fix it really any harder than the effort of screencapping the error?

Mazianni:
whoopsie

PR #1303, opened 2019-07-06
Updates floor descriptions

Andrew-Fall:


PsyCommando (approved):
I mean, you literally just changed some strings.. Not sure why that would even need a review ^^;
But it looks fine.

PR #1307, opened 2019-07-08
Fixes addiction probability

NataKilar:
NataKilar cannot into statistics. Should make addiction less prevalent.


PsyCommando (approved):
looks good

PR #1308, opened 2019-07-08
Assorted bike fixes

NataKilar:
Bikes no longer use fuel when being dragged while off.
All subtypes of ethanol can now be used as fuel.
Readds icons for cruise control and engine toggle


PsyCommando (approved):
looks good

PR #1312, opened 2019-07-08
Remove smart revolver registration

Draxtheros:
Currently the revolver kill mode exists but is locked out by a program that I do not believe exists, and most people wouldn't have access to anyway, making this gun a more expensive, less effective reskin of the Perun that doesn't fire if you forget to register it every time you come out of cryo.


PsyCommando:
That's the smart variant. There's already a standard variant that doesn't have lethal lock.

Draxtheros:
I'm aware, but the smart functionality doesn't currently do anything but reduce the functionality on a gun that's already more expensive than it's non-smart variant. As far as I'm aware the program for it doesn't exist, and if it does the general consumer can't access it, so companies can't toggle the gun's lethal auths. It's effectively defunct on Persistence, unless someone fleshes it out.

PR #1315, opened 2019-07-09
Labor Exile Implants

Mazianni:
🆑 
+ 'Exile' Teleporter hub. If you've got a labor exile implant inside of you, these won't let you teleport out.
+ Labor Exile implant. These little guys can be set to self-destruct after a period of time no longer than seven days, or manually destroyed by the exile management console.

This is intended to allow for the implementation of temporary exile to a prison station, so people don't sit in a cell for days on end.
🆑


PsyCommando:
I thought brawler didn't like teleporters?

Mazianni:
Alright i'm not sure what's wrong with the compile error, here? It's something to with personal cryopods so i assume that's brawler's doing.

Brawler never told me about a dislike of teleporters and i've been going on about this for two or three days, so.

michaelchessall:
i like teleporters!!

PR #1321, opened 2019-07-10
Fixing SSautosave

PsyCommando:
* Moved all processing and announcements into SSautosave, so its not run by SSevents anymore (which probably won't work properly because all events on the map were removed.)
- For some reasons the stat line for SSautosave won't get updated anymore for some reasons..


PsyCommando:
I put it in draft because I can't figure why the stat line on the MC tab won't update for SSautosave. The time until next save shown there is only updated once. But the subsystem still saves and announces properly.

PsyCommando:
Fixed the issue with the time to next save not being displayed correctly.

PR #1336, opened 2019-07-12
Designs should now save properly (Includes important notes about saved vars list)

ingles98:
**Important Notes:**
On the official server, atom's variable "designer_unit" should no longer be saved!
Should now be saving "designer_unit_saved_id" instead.

/datum/designer_unit

- id
- title
- icon_height
- icon_width
- icon_offset_x
- icon_offset_y
- creator_ckey
- colors
- baked_image (this var is only meant to properly save and load the icon. icon_custom is dumped into this var on save, and loaded back again on load, to its former variable name. icon_custom may no longer be set to save)

GLOB.designer_system

- save variable list "designs"


PsyCommando:
There started to be some debug messages 
`ADMIN LOG: EVENT Failed to load designer_unit at "SSHAIBASSA". No numeric ID ( '' ) (LOC)`

PR #1338, opened 2019-07-13
Removes duplicate device cell entry

Draxtheros:
There were two entries for a 25w device power cell, one of which costs significantly more and printed with zero charge. This removes it from the fabs list.


PsyCommando (approved):
Looks good!

PR #1345, opened 2019-07-14
voidsuit fixes

Mazianni:
see title


PsyCommando (approved):
More material means its better :P
Anyways, looks fine.

PR #1356, opened 2019-07-16
add tormented mob

HbsRed:


code/modules/mob/living/simple_animal/hostile/tormented.dm
```diff
@@ -0,0 +1,41 @@
...
+	speed = 4
+
+/mob/living/simple_animal/hostile/tormented/Allow_Spacemove(var/check_drift = 0)
+	return 1
+
+/mob/living/simple_animal/hostile/tormented/FindTarget()
+	. = ..()
+	if(.)
+		audible_emote("growls at [.]")
+    playsound = 'sound/hallucinations/growl1.ogg'
```
PsyCommando: That doesn't do anything.

PsyCommando (commented):
Looks alright. Just it won't make the growling sound, because you're not actually calling playsound, but affecting it for some reasons

PR #1360, opened 2019-07-21
dirts annoying

HbsRed:

### DIFF
--- a/code/game/turfs/simulated.dm
+++ b/code/game/turfs/simulated.dm
@@ -89,12 +89,12 @@
 	tracks.AddTracks(bloodDNA,comingdir,goingdir,bloodcolor)
 
 /turf/simulated/proc/update_dirt()
-	dirt = min(dirt+0.5, 101)
+	dirt = min(dirt+0.1, .5)
 	var/obj/effect/decal/cleanable/dirt/dirtoverlay = locate(/obj/effect/decal/cleanable/dirt, src)
 	if (dirt > 50)
 		if (!dirtoverlay)
 			dirtoverlay = new/obj/effect/decal/cleanable/dirt(src)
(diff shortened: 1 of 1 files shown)


code/game/turfs/simulated.dm
```diff
@@ -89,12 +89,12 @@
...
 
 /turf/simulated/proc/update_dirt()
-	dirt = min(dirt+0.5, 101)
+	dirt = min(dirt+0.1, .5)
 	var/obj/effect/decal/cleanable/dirt/dirtoverlay = locate(/obj/effect/decal/cleanable/dirt, src)
 	if (dirt > 50)
 		if (!dirtoverlay)
 			dirtoverlay = new/obj/effect/decal/cleanable/dirt(src)
-		dirtoverlay.alpha = min((dirt - 50) * 5, 255)
+		dirtoverlay.alpha = min((dirt - 50) * 5, 120)
```
PsyCommando, line 97: Why cap the alpha to 120 exactly?

code/game/turfs/simulated.dm
```diff
@@ -89,12 +89,12 @@
 	tracks.AddTracks(bloodDNA,comingdir,goingdir,bloodcolor)
 
 /turf/simulated/proc/update_dirt()
-	dirt = min(dirt+0.5, 101)
+	dirt = min(dirt+0.1, .5)
```
PsyCommando, line 92: Why? 
Also, write 0.5, not .5. Otherwise dm is gonna shit itself.

PsyCommando (changes requested):
It might be smarter to actually just reduce the dirt mobs generates, rather than doing those really weird and random changes to how the dirt overlays are displayed?

PR #1361, opened 2019-07-23
Fountain animations :)

rodney-mckay1:
Adds animations for fountain, fountain_g, and fountain_g_tb.


rodney-mckay1:
Need a way to construct a fountain IC.

PsyCommando:
there is. Stone blocks.

PR #1363, opened 2019-07-24
Update world.dm - Increases view range by two

UndeadNappist:
The city is designed for a larger view range in mind. It's either do this or re-arrange how the city looks in it's entirety.

This will also make it so people can see each other across the room instead of being inexplicably nearsighted.

By popular demand.

### DIFF
--- a/code/world.dm
+++ b/code/world.dm
@@ -7,7 +7,7 @@
 	mob = /mob/new_player
 	turf = /turf/space
 	area = /area/space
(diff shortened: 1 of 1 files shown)


PsyCommando:
I doubt that's really a good idea to change the view range out of the blue. We're having server perfs issues right now, and adding view range to iterate through probably won't help. It might be smarter to postpone this until we made things more stable.

PR #1365, opened 2019-07-25
Workout gear for the fat spessman

HbsRed:


PsyCommando:
Everything in that commit already exists..
* You can make a punching bag with plastic sheets.
* You can dismantle weightlifting machines. 
* You can make all the balls.
* Only thing you couldn't make was the basketball hoop, but its hacky as hell, and I don't know if it'll work too well.

PR #1367, opened 2019-07-28
Update digitalwarrant.dm

rhicora:
- Makes Judges able to add warrants, and all able to view them.


PsyCommando:
I think brawler removed all the warrant stuff?
Does it even do anything at all now?

PR #1371, opened 2019-07-31
autosave to status

HbsRed:


HbsRed:
I've totally pr'd unknowing if this is even the way to add something to the status bar. I hope it works

PR #1373, opened 2019-07-31
Autosave status

HbsRed:

### DIFF
--- a/code/modules/mob/mob.dm
+++ b/code/modules/mob/mob.dm
@@ -693,6 +693,7 @@
 			stat("Local Time", stationtime2text())
 			stat("Local Date", stationdate2text())
 			stat("Server Uptime", roundduration2text())
+			stat("Autosave Timer"), "Next autosave in [round(time_diff/(1 MINUTE), 0.1)] minutes."
 		if(client.holder || isghost(client.mob))
 			stat("Location:", "([x], [y], [z]) [loc]")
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/mob.dm
```diff
@@ -693,6 +693,7 @@
 			stat("Local Time", stationtime2text())
 			stat("Local Date", stationdate2text())
 			stat("Server Uptime", roundduration2text())
+			stat("Autosave Timer"), SSautosave.time_next_save
```
PsyCommando, line 696: The parenthesis at the end is missing. Also, that would display time in deciseconds, which would be a huge meaningless number to most people.

PsyCommando (changes requested):
You didn't compile this or even test it did you? :/

HbsRed:
See thing is.. i follow brawlers good ol advice and hes like yup sounds
great then sha bang. It no work pepe but you see what im trying to do yeah?

PsyCommando:
> 
> See thing is.. i follow brawlers good ol advice and hes like yup sounds great then sha bang. It no work pepe but you see what im trying to do yeah?
> […](#)

Well I mean, my main concern is having the code work, so the server works XD Which is why I'm requesting changes on this pr so it can still work if this is merged XD

PsyCommando:
I guess I'll close this one, since you don't seem interested in fixing the PR.

PR #1380, opened 2019-08-16
Fixes some missing belt icons

Draxtheros:
Several security belts had incorrect icon states due to missing sprites in the DMI. Now they'll at least be visible.


PsyCommando (changes requested):
I straight up removed those, because they had no sprites, and even in bay's repo I couldn't find a sprite for them at all. But of course it got reverted because.
There's no point really in putting those back in since they don't function differently than the other belts, and they don't even have a proper sprite anymore. 

I recommend just commenting the lot of them out and leaving it at that. Otherwise it'll become confusing.

Draxtheros:
Aaand they're gone

PR #1394, opened 2019-08-22
restore the missing map z-levels

PsyCommando:
Brawler removed the unit test z-level so I put it back in so unit tests can work again.


PsyCommando:
Would go well with #1423

PR #1421, opened 2019-08-25
RE-Somi

Draxtheros:
New species, except not. This ports Resomi back into the code, with lore in the works for it's final inclusion.


code/modules/species/species.dm
```diff
@@ -35,6 +35,7 @@
 	var/tail_blend = ICON_ADD
 	var/tail_hair
 
+
```
PsyCommando: Would be good to avoid commiting whitespace changes.

PsyCommando (changes requested):
Looks fine! 
Can you just add proper unit tests for the resomi too?

Draxtheros:
Typo in the unit tests fixed, as well as the extra white space

PsyCommando (approved):

PsyCommando:
Alright, looks good to me then!

PR #1422, opened 2019-08-25
More docking pad updates.

biscuitWizard:
Added the UI changes for the new beacon.


biscuitWizard:
I updated the template so the new sizes are actually selectable now.

PR #1423, opened 2019-08-26
Fixes Travis (somewhat)

Bxil:


maps/nexus/nexus.dm
```diff
@@ -7,7 +7,7 @@
 	#include "../../code/modules/lobby_music/comet_haley.dm"
 	#include "../../code/modules/lobby_music/lysendraa.dm"
 
-//	#include "nexus.dmm"
+	#include "nexus.dmm"
```
PsyCommando: Can you re-comment that? Because of the way things work right now, we gotta keep the empty map on load, or it'll ovewrite stuff on load and create a lot of weirdness.

PsyCommando (changes requested):
Helps a bunch tbh!
Just re-comment the base nexus map, so we don't end up with weird issues on load, and it should be good.

maps/nexus/nexus.dm
Bxil: A proper map like that is needed to be included. Around 80 of the unit test failures were caused by this.
With this we are down to one, and that one failure is legitimate.

PsyCommando: Its because brawler deleted the testing z-level on the "empty" version of the map. If you include this thing, there's gonna be way worst things than travis failing.

Bxil: Done.

PsyCommando: Ok!

PsyCommando (approved):

PsyCommando:
Hopefully the test z-level will get merged eventually too...

PR #1424, opened 2019-08-26
Removing some slimes

Draxtheros:
Black and oil slimes hold a lot of griefing potential, and not much else.
Green slime, while neat in effect and concept, turn people into an utterly broken species that has no whitelist and a lot of potential for abuse, with near to no counterbalancing negatives.


PsyCommando (approved):
Looks good to me!

PR #1428, opened 2019-08-29
Defib battery fix

fernandos13:
Defib battery fix


code/game/objects/items/weapons/defib.dm
```diff
@@ -18,6 +18,9 @@
 	var/obj/item/weapon/shockpaddles/linked/paddles
 	var/obj/item/weapon/cell/bcell = null
 
+	/obj/weapon/cell/New(var/newloc, var/model)
+	ADD_SAVED_VAR(bcell)
```
PsyCommando, line 23: The indent is wrong and the type path for the new is also wrong, and it doesn't call the base class's new proc either.

PsyCommando (changes requested):
This doesn't actually fix anything. Please compile/test your stuff before making a PR.

code/game/objects/items/weapons/defib.dm
```diff
@@ -14,10 +14,13 @@
 	w_class = ITEM_SIZE_LARGE
 	origin_tech = list(TECH_BIO = 4, TECH_POWER = 2)
 	action_button_name = "Remove/Replace Paddles"
-
+	var/bcell = null
```
PsyCommando, line 17: Well, you don't need to define the bcell var again.

code/game/objects/items/weapons/defib.dm
```diff
@@ -14,10 +14,13 @@
...
 	var/obj/item/weapon/shockpaddles/linked/paddles
 	var/obj/item/weapon/cell/bcell = null
 
+/obj/item/weapon/defibrillator/New()
```
PsyCommando, line 21: Just add 
`..()`
On a new line after 
`/obj/item/weapon/defibrillator/New()`
so that it calls the base New() proc of the parent classes.

PsyCommando (changes requested):
Just remove the double declaration for the bcell var, and call the base New proc in the defib to fix the PR.

PR #1429, opened 2019-08-29
[Recovery PR] Make cryopods a whole lot simpler, and work

PsyCommando:
* Reduced the copypasta
* Removed redundant vars and used the existiing spawn type var instead.
* Made initial character spawning fill up all the vars it really should always have.
* Made the ghostize() actually return the ghost mob when its called. Which fix many things.
* Also, save the personal cryo coordinates that weren't being saved at all
* Blood type is properly set in the DNA now


code/game/machinery/cryopod.dm
```diff
@@ -8,6 +8,7 @@
 	dir = WEST
 	req_access = core_access_command_programs
 	circuit_type = /obj/item/weapon/circuitboard/cryopod
+	atom_flags = ATOM_FLAG_NO_TEMP_CHANGE | ATOM_FLAG_CLIMBABLE
```
Andrew-Fall: Won't CLIIMABLE prevent climbing inside?

PsyCommando: I guess the check for climbing stuff happens a while after climbing into the thing. Because I didn't notice that when testing.

PsyCommando: Removed the flag anyways.

PR #1430, opened 2019-08-30
Rebalance device cells, egun shot costs

Draxtheros:
Previously there were two kinds of device cells, rated 25 and 100 each. They cost absolutely nothing to make, and were useless to boot due to most guns getting, at most, five shots from the 'high end' cell. This rebalances each to 100, 200, and adds a new cell at 300 power rating while increasing the cost for each significantly. Energy firearms have had their shot cost rebalanced to give them a standard shot progression of 5, 10, 15 with each cell, putting them more in line with ballistics without having to game the system (prior to this, you could make a 300 power rated device cell by crafting a gun and removing the cell). Guns with different shot progressions have been accounted for. For a non exhaustive list of examples, the laser cannon (3, 6, 9) and the laser sniper (2, 4, 6) each have a higher or lower shot charge cost to account for the new cell system, while retaining their original shot number at the mid-level cell.


PsyCommando (approved):
Looks good,
But there's probably more to do to completely fix energy weapons.

Draxtheros:
Probably, but I'll fix them as they get pointed out.

PsyCommando:
Yeah, the main issue is how energy weapons were meant to create their own internal cell with custom charge depending on the capacity indicated in the parent weapon. Custom cells weren't considered before we changed it I think.

Draxtheros:
Yeah, that was the largest issue. The cell charge numbers I've got here are based around that, 300 is the largest weapon cell I could find, so this will balance the system without leaving any broken, overpowered cells in the server. As it is, most weapons should spawn with a 200 power cell with these changes

PR #1431, opened 2019-08-30
Update subtypes.dm

Zembaflemba:
Made slime breeding slightly less fucky by allowing all slimes to multiply into the same color and removing grey from all colors except grey itself.


Zembaflemba:
pls am do merge because breeding slimes is hell now i am to thank

PsyCommando (approved):
Seems fine to me. Looking at the bay stuff it seems pretty similar. I don't know when that weird nerf happened..

PR #1432, opened 2019-08-30
Nanopaste for limbs

fernandos13:
this should let people use nanopaste as a bruise pack (BUT ONLY for damages under 15)


code/game/objects/items/stacks/medical.dm
```diff
@@ -12,6 +12,7 @@
 	var/heal_burn = 0
 	var/animal_heal = 3
 	var/apply_sounds
+	var/damage_amount
```
PsyCommando, line 15: There's no reasons for you to put this here.

code/game/objects/items/stacks/nanopaste.dm
```diff
@@ -46,3 +46,17 @@
 				use(1)
 				user.visible_message("<span class='notice'>\The [user] applies some nanite paste on [user != M ? "[M]'s [S.name]" : "[S]"] with [src].</span>",\
 				"<span class='notice'>You apply some nanite paste on [user == M ? "your" : "[M]'s"] [S.name].</span>")
+		if(S && BP_IS_ROBOTIC(S) && S.hatch_state == HATCH_CLOSED)
+			if(damage_amount >= 15)
```
PsyCommando, line 50: Why do you use that var that doesn't even exist in this context to check the damages on the limb? You probably didn't compile this, right?

PsyCommando (changes requested):
You probably want to redo this one. Its not doing anything like you claim it to do. And it actually doesn't do anything at all.
Also usually, you probably want to put object specific interactions in attackby of the receiving object.

PsyCommando:
Also you didn't even compile this before submitting the change. It fails when compiled :

```
code/game/objects/items/stacks/nanopaste.dm:50:error: damage_amount: undefined var

code/game/machinery/cryopod.dm:346:warning: occupant_ref: variable defined but not used

persistentss13.mdme.dmb - 1 error, 1 warning (8/31/19 12:57 pm)
```

PR #1435, opened 2019-09-01
Resomi fixes

Draxtheros:
The language was not included, nor were names set up generate properly based on the species language. Additionally, some clothes that were missing their in-world icon have been fixed.


code/modules/species/station/resomi.dm
```diff
@@ -85,6 +85,17 @@
 		/datum/mob_descriptor/height = -2,
 		/datum/mob_descriptor/build = -1
 		)
+		
+	available_cultural_info = list(
```
PsyCommando, line 89: I'd add the other basic tags too. Just in case. Because culture tags are super finicky.

PsyCommando (changes requested):
Looks good! Just add the other standard `available_cultural_info` for the other default tags. Because, the thing is very finicky, and it'll make it compatible with standard bay too.

code/modules/species/station/resomi.dm
Draxtheros: It should be fine, the Human entry also only has the two, with the rest defaulting to whatever is in the base species.dm

PsyCommando: Alright!

PsyCommando (approved):

PR #1438, opened 2019-09-10
More Resomi Fixes

Draxtheros:
Resomi now start with sandals, with which to protect their feet from the infinite piles of glass people seem to like to amass.
Resomi no longer have the Bald hair style, because their ears are included in their hair, you monster.
Resomi are now slightly less buff visually, beause I don't know, reasons.


PsyCommando (approved):
Looks good!

Draxtheros:
Slid in the Resomi modkit onto a fab, since I'd noticed I hadn't done that yet

PR #1444, opened 2019-10-04
Improve cyborg saving/loading + misc cyborg things

PsyCommando:
* Cyborg modules and etc save.
* Made some important robot/IPC part buildable.
* Reworked matter synth to require decompiled matter to recharge.
* Reworked cyborg ids to take into account the brainmob's id.
* Changed some of the presets for the robot types
* Added circuit for robot cryopod


Andrew-Fall:
Conflict, and I think regular cryopods should allow robots instead of having the separate one for them

PsyCommando:
Fixed, and yeah, but its really just an aesthetic thing. And I'm not gonna touch the cryopod again until we can merge my cryopod stuff in, since the state they're currently in is kinda weird.

PR #1446, opened 2019-10-04
Add back some carp spawns and other mobs spawns

PsyCommando:


code/modules/maze_map/maze_map.dm
```diff
@@ -189,7 +196,15 @@ GLOBAL_LIST_EMPTY(maze_map_data)
...
 	W_connect = 7
-	monster_types = list(/mob/living/simple_animal/hostile/tormented, /mob/living/simple_animal/hostile/creature) // types of monsters that will occur on this map.
+	monster_types = list(
+		/mob/living/simple_animal/hostile/tormented, 
+		/mob/living/simple_animal/hostile/voxslug, 
+		/mob/living/simple_animal/hostile/scarybat, 
+		/mob/living/simple_animal/hostile/creature,
+		/mob/living/simple_animal/hostile/carp/pike, 
+		/mob/living/simple_animal/hostile/vagrant,
+		/mob/living/simple_animal/hostile/bear,
```
Andrew-Fall, line 206: I suggest removing the bears, they don't really fit

PsyCommando: Well, I mean, they're literally space bears. And also the only reason I'm putting those back in is because people can't get carp meat, leather and stuff from those anymore because there's just a single enemy mob type now.

PR #1451, opened 2019-11-02
Adds Sycorax Station as a temporary map

needanewgithub:
Map was made over a day or two to create a temporary map to give players something to do until a full feature-complete map is created.

This document sets out the general design: https://drive.google.com/file/d/1uhpIuBR84Xd3-i2XRbIGLEB3ZTaVaz4g/view?usp=sharing


needanewgithub:
Large changes were made to map, map now consists of two separated ships. Ship one, the "Phobetor" is mostly done, ship two the "Xerxes" needs to be redesigned and finished.

My idea is that one ship would serve as the headquarters of the government, while the other would be for an alternative and harder corporate spawn for players who aren't interested in directly interacting with the government, or for characters who wish to flee from the government.

needanewgithub:
Additionally, the titular "Sycorax Station" will be a hellstation-esque spawn with minimal infrastructure to allow players to establish themselves on the map outside of the two spaceships. Zlevels that can be easily built on aside from these won't be accessible from the start.

